PR #48, opened 2022-01-24
Add support for running all the SAP roles on a single server

KimForss:
## Problem
The current playbooks do not allow for collapsing all of the roles to a single server

## Solution
Add logic to enable the single server deployment by
- ensuring that the instance ID for PAS is different from the default 00
- creating the installation media volume on the standalone server instead of on the SCS server

## Tests
Run an deployment with 
enable_app_tier_deployment = false

## Notes


hdamecharla (approved):
Hi @KimForss , Approved #48 👍🏾 🙏🏾

PR #83, opened 2022-02-11
BOM folder path location change

bajajyukta7:
## Problem
BOM folder path was getting printed in the middle of the output making it difficult to read by the user.

## Solution
Changed the location of code to print the bom folder path at the end of output.

## Tests
Validated the playbook by download/upload sapbits for S4 hana v10 bom.
Screenshot 2022-02-11 174620

## Notes


hdamecharla (approved):
@bajajyukta7  Approved #83

PR #124, opened 2022-04-01
Create S/4HANA 2021 Bill of materials

hdamecharla:
## Problem
We do not have a S/4HANA 2021 Bill of materials

## Solution
Create BOM based on [BOM prep process](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/automation-bom-get-files)

## Notes
This is the initial version and will be modified as we move through the testing process


KimForss (approved):
Looks good to me

PR #136, opened 2022-04-07
Push the S4HANA 2021 bom and tempates for distributed installations.

hdamecharla:
TODO: Need to prepare templates for HA deployments.

## Problem
Enhance #124 and provide the templates for distributed installation of S/4 HANA 2021 ISS

## Solution
Create Jinja templates for use with the solution

## Notes
TODO: Need to prepare templates for HA deployments.


KimForss (approved):
Looks good to me @hdamecharla

PR #150, opened 2022-04-25
Install flags in the progress directory don't need root privileges

hjstam:
## Problem
The install flags in the workspace progress directory don't need root privileges. The progress directory in the workspace directory is from the user (azureadm by default on the deployer) running the Ansible command.

## Solution
Add become false to delegated_to localhost tasks.

## Notes


KimForss (approved):
Looks good to me

PR #153, opened 2022-05-03
Deploy zone-redundant load balancers by default

hjstam:
## Problem
As of AzureRM version 3.0 the frontend_ip_configurations of Azure Load Balancers are deployed without zone redundancy. After creation the zone(s) attribute is returned as a Null field and the Azure portal shows an empty select field.

The zones attribute is optional (and isn’t defaulted by Azure): omitting this field (or specifying a value of null) will deploy this without any Zones.

## Solution
As the default behaviour before let to Zone-Redundant load balancers we can simply set the zones of the frontend_ip_configuration to all Availability Zones.

## Notes
Sources:
- https://www.hashicorp.com/blog/terraform-azurerm-3-0-brings-enhanced-azure-function-support
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb#zones


KimForss (commented):
Thanks for this

PR #156, opened 2022-05-08
Fix Ansible Lint violations

hjstam:
## Problem
Ansible Lint is flagging 16 violations

## Solution
Fix these 16 violations

## Notes


KimForss (approved):
Thank you

PR #164, opened 2022-05-15
Add hdbadm_uid and sapsys_gid to HANA installation templates

hjstam:
## Problem
The HANA installation isn't using the hdbadm_uid and sapsys_gid variables from the ansible-input-api.yaml file.
The db_pwd variable is obsolete. The main_password variable can be used directly in the HANA installation template.

## Solution
Use the hdbadm_uid and sapsys_gid variables in the HANA installation templates.
Remove db_pwd variable from hdb-install main task.


KimForss (approved):

KimForss:
Thanks @hjstam

PR #168, opened 2022-05-30
Merging Private Preview

KimForss:
## Problem
Add the updates for the next release

Support for Azure NetApp files
 - HANA data, log, shared
 - Mounting /usr/sap from Azure NetApp files

Auto configuration of self hosted agent
Removal of sas-token dependency
Support for RBAC on workload zone key vault
Improved Multi-sid support

Better configurability for brownfield deployments of shared storage
 - Private endpoints

Imprived error messages in pipelines
Support for multiple deployers
Support for providing an additional IP per NIC
Default sizing configuration for the supported E series VMs


hdamecharla (approved):
Hi @KimForss , #168 Looks good to me. Approved. 🙏🏾

PR #171, opened 2022-06-03
Secrets in scripting require quotes

ErickSegaar:
## Problem
When running the pipeline following the described documentation (creating the library SDAF-MGMT with ARM_CLIENT_SECRET) the shell scripts will fail with the following:
```
/home/vsts/work/_temp/547efcbd-8b04-4720-a04d-61af9a172d22.sh: line 44: syntax error near unexpected token `)' 
/home/vsts/work/_temp/547efcbd-8b04-4720-a04d-61af9a172d22.sh: line 44: ` if [ ! -n *** ]; then' 
```

## Solution
Adding quotes to surround the ARM_CLIENT_SECRET to prevent the value of the secret impacting the script

## Tests
Run the 01-deploy-control-plane.yaml pipeline

## Notes
It only is a problem when inside the scripting

### DIFF
--- a/deploy/pipelines/01-deploy-control-plane.yaml
+++ b/deploy/pipelines/01-deploy-control-plane.yaml
@@ -98,7 +98,7 @@ stages:
                   echo "##vso[task.logissue type=error]Variable ARM_CLIENT_ID was not defined."
                   exit 2
                 fi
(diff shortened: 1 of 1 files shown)


KimForss (approved):
Thanks, good catch

PR #173, opened 2022-06-03
Add variable for BOM processing without root privileges on the deployer/controller node

hjstam:
## Problem
It's currently not possible to use the BOM processing on a Ansible controller node without root privileges.

## Solution
By overriding the download_directory to a location where the executing user has privileges the become options can be set to false. The BOM processing become arguments can now be managed with the bom_processing_become variable.


KimForss (approved):
Thanks - looking good

PR #175, opened 2022-06-08
use agent and remove unused pool variable

ErickSegaar:
## Problem
The documentation describes the use of a POOL variable in the library to use your hosted agent. This POOL variable is actually never used and the vmimage is set hard for the pipeline.

image

## Solution
Replaced the hard reference with the $(Agent) that is being used by other parts of the code and is defined in your second library

## Tests
Run the 02 pipeline

## Notes
Should also be removed from the documentation cause it is doing nothing [documentaion](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/automation-configure-devops#common-variables)

### DIFF
--- a/deploy/pipelines/01-deploy-control-plane.yaml
+++ b/deploy/pipelines/01-deploy-control-plane.yaml
@@ -228,6 +228,7 @@ stages:
               ARM_TENANT_ID:           $(ARM_TENANT_ID)
               TF_VAR_agent_pat:        $(PAT)
               TF_VAR_agent_ado_url:    $(System.CollectionUri)
+              TF_VAR_agent_pool:       $(POOL)
             failOnStderr:              false
 
 ...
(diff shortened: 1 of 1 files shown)


KimForss:
Hi @ErickSegaar 

The POOL parameter is used for the automatic configuration of the deployer to automatically register the deployer as an agent.

https://github.com/Azure/sap-automation/blob/dcd55edef8754a893e12c2f839ca5c3798415da0/deploy/terraform/terraform-units/modules/sap_deployer/templates/configure_deployer.sh.tmpl#L484

I can make that more visible in the documentation.

//Kimmo

ErickSegaar:
> 

Ok, missed that one when I searched for it. Shall I update the PR with the POOL environment variable back in it. But remain the $(agent) for the running of the pipeline?

KimForss:
Yes please
//Kimmo

ErickSegaar:
Ok cleaned it up

KimForss (approved):

PR #176, opened 2022-06-10
Bug/fix pep subnet

UncleBats:
## Problem
When using a private endpoint the need for subnet exclusion is not needed. Especially when you need to use a managed vnet from a CCEO which locks the resource. Then you won't be able to apply the virtual_network_subnet_ids (the lock will prevent that) 

## Solution
Whenever you enable the private enpoint the need for allow networks to the resource is not needed

## Tests
Run pipeline 02 once with private enpoint enabled => expectation a deployment with pep created and no networks added to the resources
Run pipeline 02 once without private endpoint => expectation vnet added to the network rules

## Notes
This will probably be in more resources, like also disabling the default allow for keyvaults but that will be a separate PR


KimForss (approved):
Thanks @ErickSegaar

PR #177, opened 2022-06-13
Set default network rule action based on configuration

ErickSegaar:
## Problem
When you assign a private endpoint, IP or a subnet the default_action should be set to deny, otherwise, they will still be reachable from public networks. 

This is explained in the documentation
image

## Solution
Made it conditional based on the already existing variables

## Tests
Run the 02 pipeline

## Notes


KimForss (approved):
Thanks

PR #179, opened 2022-06-13
Bugfixes for ha cluster setup

Rickmarges:
## Problem
- The variables 'primary_instance_ip_db' and 'secondary_instance_ip_db' are undefined in 4.0.1.3
- Task file '2.10.1.1.yaml' could not be found in task '2.10.1.yaml'
- The directory '{{ tmp_directory }}/{{ sid | upper }}' does not exist, resulting in installation using the wrong temp dir
- The wait for async task wails due to the job not existing anymore
- The app install uses the wrong virtual hostname during install
- Post SCS HA install fails due to delegation from ERS vm
- Pacemaker resource fails due to filesystem device not being available
- When not using a deployer the creation of the keyvault fails due to there not being a deployer subnet

## Solution
- Create the variable in playbook_04_00_01_hana_hsr.yaml
- Provide path for task file '2.10.1.1.yaml'
- Create a task to ensure the temp directory is created
- Removed the wait task. It's not needed due to the poll being present
- Update variable to use the correct hostname
- Remove the delegation and run the task directly from the Ansible controller to the SCS vm
- Make the device name, used for creating the resource, variable
- Use the Terraform compact function to ensure the list is empty and not used during creation


KimForss:
Hi @Rickmarges I just pushed a new branch "experimental" - we want to start using that for pre-release code as it resonates better with the open source philosophy.

Can you target the PR against that branch?

Thanks,
Kimmo

Rickmarges:
Hi @KimForss 

Done :)

deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
```diff
@@ -140,20 +146,10 @@
       async:                           7200
       poll:                            60
 
-    - name:                            "DBLoad Install -Wait for asynchronous job to end"
```
KimForss, line 143: We need this to be here - Azure DevOps tasks will otherwise fail if they run for longer than 60 minutes

Rickmarges: But due to the poll and async in the previous task it stays there until it's either finished, or if 2 hours have passed.
In our case it's finished within those 2 hours and that results in the next task (the one I removed) failing, because the job cannot be found.

As far as I'm aware either the poll in the async task needs to be 0, or the wait needs to be removed

KimForss: Hi, I've based mine on this: https://docs.ansible.com/ansible/latest/user_guide/playbooks_async.html 

and have been using it without any issue
image

Without it our Azure DevOps tasks fail - I assume that is because if there is no action on the main thread

Rickmarges: Hi Kimmo I just ran some tests locally with the following code: 

```yaml
- name: Async sleep for 15 seconds
  ansible.builtin.command: sleep 15
  register: results
  async: 60
  poll: 10

- name: check status (wait task)
  ansible.builtin.async_status:
    jid: "{{ results.ansible_job_id }}"
  register: job_result
  until: job_result.finished
  retries: 90
  delay: 10
  when: results.ansible_job_id is defined
 ```
 With the poll in the first task i get the following output:
 ```
 TASK [Async sleep for 15 seconds] *************************************************************************************************************************************************************************************************
ASYNC POLL on localhost: jid=317019997408.17994 started=1 finished=0
ASYNC OK on localhost: jid=317019997408.17994
changed: [localhost]

TASK [check status] ***************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"ansible_job_id": "317019997408.17994", "attempts": 1, "changed": false, "finished": 1, "msg": "could not find job", "results_file": "/home/aa618475/.ansible_async/317019997408.17994", "started": 1, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}

PLAY RECAP ************************************************************************************************************************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

The async task is already finished when we get to the wait task, so there is no job to wait for anymore.
The other option is to set the poll to 0 to go straight to the wait task. That will result in the following output:

```
TASK [Async sleep for 15 seconds] *************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [check status] ***************************************************************************************************************************************************************************************************************
FAILED - RETRYING: [localhost]: check status (90 retries left).
FAILED - RETRYING: [localhost]: check status (89 retries left).
changed: [localhost]

PLAY RECAP ************************************************************************************************************************************************************************************************************************
localhost                  : ok=2    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

KimForss: ok - let me run the same using our pipelines

Rickmarges:
I will close this PR and open a new one to have less commits

PR #180, opened 2022-06-13
Add custom umask to install commands

Rickmarges:
## Problem
There is no possibility to set the usmask during installation.

## Solution
The install commands now have a default umask of 022 and the possibility to overwrite this by using the 'custom_umask' variable


KimForss:
@Rickmarges this is a great idea. 

However the PR has some of the indentation's wrong. 
I also removed some of the old code "refactor folder".

Can you either update the PR or maybe easier to do a new one. 

//Kimmo

Rickmarges:
I will create a new pull request

PR #183, opened 2022-06-14
SSH access between nodes isn't required for pacemaker on RHEL

hjstam:
## Problem
SSH root access between nodes in Enterprise environments usually isn't permitted. According to the docs this is however a requirement for crm_clustering on SUSE but it isn't for RHEL.

## Solution
Only generate and distribute SSH private/public keys for SUSE systems and not RHEL.

## Notes
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-rhel-pacemaker


KimForss (approved):
Looks good - will run a test on my end as well

PR #184, opened 2022-06-15
Fix all ansible-lint failures

Rickmarges:
## Problem
All builds are failiing, which results in worsening code checks

## Solution
I fixed all ansible-lint violations. After this all checks should succeed and future code can be checked easier

## Tests
Ran: `ansible-lint -v --force-color` just as GitHub actions does


KimForss (approved):
Thanks

KimForss:
@Rickmarges Thanks - this will help a lot

KimForss:
@Rickmarges I had to revert it as it broke our naming/indentation. 

We align the task name and the parameters on column 40 for readability

Rickmarges:
@KimForss i get that it's for readability, but as far as Ansible cares that's not correct.
Ansible always want the have the parameter straight after the colon.

The one thing I don't get is how this can break the naming?
And if it does break the naming, shouldn't that be fixed?

Rickmarges:
@KimForss I also noticed I've created some issues due to serach and replace on items. 
One example of this is: `db_sidpper` instead of `db_sid | upper`.
I will look into this and fix this. But I still think this code has to be merged evntually

KimForss:
Sorry, my bad - it does not break naming. But one of our criteria is readability, and having the task name and the parameters aligned on position 40 makes it easier
We run the linter using

exclude_paths:
  - deploy/pipelines
  - deploy/ansible/BOM-catalog/archives
  - deploy/terraform
  - .github

warn_list:
  # FIXME: Experimental violations should be fixed in separate PR
  - experimental

skip_list:
  # Ignored because there is no easier way to do
  - empty-string-compare

  # Following are ignored because of readability:
  - line-length
  - comments
  - comments-indentation
  - colons
  - commas
  - braces
  - hyphens
  - role-name
  - var-naming
  - no-changed-when

  #csh is not supported in command
  - command-instead-of-shell

Rickmarges:
Could you provide me with the criteria your code needs to comply with?
Maybe I can fix the linting while matching those

Rickmarges:
Okay, I will see if i can fix the issues. But there are also some which are quite easy to fix that makes sense, like the space at the beginning of a comment. I don't see why that should be excluded instead of fixed

KimForss:
I agree, we added some of the exceptions as the first lint result list was killing us 😀

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Rickmarges:
@KimForss it might be a good idea to also pin the ansible-lint version in the GitHub action, as the version changes constantly. With the constantly changing version the checks changes a lot and makes it almost impossible to create a steady "build"

KimForss:
I agree, which version should we pin on?

Rickmarges:
@KimForss to me it doesn't matter which version. I'd suggest to just pick the latest version that is available at this point. Then fix the linting issues and you should have a successful build for the upcoming period

PR #186, opened 2022-06-15
use env vars instead of pipeline due to secret not working

ErickSegaar:
## Problem
When running the pipeline following the described documentation (creating the library SDAF-DEV with ARM_CLIENT_SECRET) the shell scripts will fail with the following:
```
--- Validations --- 
/home/vsts/work/_temp/797b59ab-def3-4170-837f-c38fd8a80ba5.sh: line 15: syntax error near unexpected token `)' 
/home/vsts/work/_temp/797b59ab-def3-4170-837f-c38fd8a80ba5.sh: line 15: ` if [ ! -n *** ]; then' 
##[error]Bash exited with code '2'. 

```

## Solution
Using the environment variables for shell scripts is advised by the documentations and resolves the issue we have for the  ARM_CLIENT_SECRET to prevent the value of the secret impacting the script

## Tests
Run the02-sap-workload-zone.yaml pipeline

## Notes
It only is a problem when inside the scripting
same as PR #171

### DIFF
--- a/deploy/pipelines/02-sap-workload-zone.yaml
+++ b/deploy/pipelines/02-sap-workload-zone.yaml
@@ -67,7 +67,7 @@ stages:
                   echo "##vso[task.logissue type=error]Variable ARM_CLIENT_ID was not defined."
                   exit 2
                 fi
-                if [ ! -n $(ARM_CLIENT_SECRET) ]; then
+                if [ ! -n $ARM_CLIENT_SECRET ]; then
                   echo "##vso[task.logissue type=error]Variable ARM_CLIENT_SECRET was not defined."
                   exit 2
                 fi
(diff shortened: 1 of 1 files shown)


KimForss (approved):

ErickSegaar:
@KimForss Now you require the PR to go into the experimental branch, but how to go from there? what is your timeline for main?

KimForss:
Plan is to do PRs every two weeks from experimental to main

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

ErickSegaar:
@KimForss ok then we can plan for that, however I just seen a change on main ([link](https://github.com/Azure/sap-automation/commit/31db7ce3df2974624f964f84fac180b1fcf257b9)) coming through directly. So how do you manage that. When you want something directly in main and when in experimental. I was under the impression that you wanted everything in experimental and you do a PR once every 2 weeks into main. So pls explain to help me help you

ErickSegaar:
Can you complete the PR merge, I will change the previous 01 pipeline as well where we used the quotes to use the environment vars aswell

ErickSegaar:
Added the modifications to control-plane aswell and included the system where I found the same

PR #188, opened 2022-06-15
Fix ansible-lint errors

Rickmarges:
## Problem
All builds are failiing, which results in worsening code checks

## Solution
I fixed all ansible-lint violations. After this all checks should succeed and future code can be checked easier

## Tests
Ran: `ansible-lint -v --force-color` just as GitHub actions does

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -3,6 +3,7 @@ exclude_paths:
   - deploy/ansible/BOM-catalog/archives
   - deploy/terraform
   - .github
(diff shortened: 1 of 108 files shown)


Rickmarges:
@KimForss Here is the updated version of the fixes for ansible-lint. 

I had to change the .ansible-lint file a little bit because of the new version, but it's mostly the same.
Only removed a few which aren't necessary anymore, and added 2 that are

KimForss (approved):

PR #190, opened 2022-06-16
Add custom umask

Rickmarges:
## Problem
There is no possibility to set the usmask during installation.

## Solution
The install commands now have a default umask of 022 and the possibility to overwrite this by using the 'custom_umask' variable


KimForss (approved):
Thanks, there are a few other places where this is needed - I'll add them manually once this is merged

PR #191, opened 2022-06-16
Fix hacluster

Rickmarges:
## Problem
- The variables 'primary_instance_ip_db' and 'secondary_instance_ip_db' are undefined in 4.0.1.3
- Task file '2.10.1.1.yaml' could not be found in task '2.10.1.yaml'
- The directory '{{ tmp_directory }}/{{ sid | upper }}' does not exist, resulting in installation using the wrong temp dir
- The wait for async task wails due to the job not existing anymore
- The app install uses the wrong virtual hostname during install
- Post SCS HA install fails due to delegation from ERS vm
- Pacemaker resource fails due to filesystem device not being available
- When not using a deployer the creation of the keyvault fails due to there not being a deployer subnet

## Solution
- Create the variable in playbook_04_00_01_hana_hsr.yaml
- Provide path for task file '2.10.1.1.yaml'
- Create a task to ensure the temp directory is created
- Removed the wait task. It's not needed due to the poll being present
- Update variable to use the correct hostname
- Remove the delegation and run the task directly from the Ansible controller to the SCS vm
- Make the device name, used for creating the resource, variable
- Use the Terraform compact function to ensure the list is empty and not used during creation

### DIFF
--- a/deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
+++ b/deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
@@ -52,7 +52,13 @@
   ansible.builtin.file:
     path:                              "{{ dir_params }}"
     state:                             directory
-    mode:                              '0755'
+    mode:                              0755
+
+- name:                                "DBLoad: Create temp directory for sid"
+  ansible.builtin.file:
+    path:                              "{{ tmp_directory }}/{{ sid_to_be_deployed.sid | upper }}"
+    state:                             directory
+    mode:                              0755
 
 - name:                                "DBLoad: - Include 3.3.1-bom-utility role"
   ansible.builtin.include_role:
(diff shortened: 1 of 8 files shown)


deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
```diff
@@ -141,20 +147,10 @@
       async:                           7200
       poll:                            60
```
KimForss, line 148:
```suggestion
      poll:                            0
```

deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
```diff
@@ -141,20 +147,10 @@
...
 
-    - name:                            "DBLoad Install -Wait for asynchronous job to end"
```
KimForss, line 144: Please bring this back

    - name:                            "DBLoad Install -Wait for asynchronous job to end"
      ansible.builtin.async_status:
        jid:                           '{{ dbload_results.ansible_job_id }}'
      register:                        job_result
      until:                           job_result.finished
      retries:                         120
      delay:                           60
      when:                            dbload_results.ansible_job_id is defined

PR #196, opened 2022-06-16
Use HANA vmname instead of secondary dnsname for single nic deployments

hjstam:
## Problem
When deploying single NIC HANA VM's the HANA_SECONDARY_DNSNAME is used in the DNS ouput.

## Solution
Use the HANA_VMNAME for single nic deployments in DNS output as this is the same for the Application VM's.

## Notes


KimForss (approved):
Thanks

PR #197, opened 2022-06-17
use environment variable for client-secret

ErickSegaar:
## Problem
When running the pipeline 10-remover-terraform.yaml the shell scripts will fail with the following:
```
--- az login --- 
/home/vsts/work/_temp/80e0e059-14c9-46c6-bd90-09a14fa250d0.sh: line 45: syntax error near unexpected token `)' 
/home/vsts/work/_temp/80e0e059-14c9-46c6-bd90-09a14fa250d0.sh: line 45: ` if [ ! -n *** ]; then' 
##[error]Bash exited with code '2'. 
```

## Solution
Using the environment variables for shell scripts is advised by the documentations and resolves the issue we have for the ARM_CLIENT_SECRET to prevent the value of the secret impacting the script

## Tests
run the 10-remover-terraform.yaml pipeline

## Notes
same as PR #171  and #186

### DIFF
--- a/deploy/pipelines/10-remover-terraform.yaml
+++ b/deploy/pipelines/10-remover-terraform.yaml
@@ -144,7 +144,7 @@ stages:
                   echo "##vso[task.logissue type=error]Variable ARM_CLIENT_ID was not defined."
                   exit 2
                 fi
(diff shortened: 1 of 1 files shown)


KimForss (approved):
Thanks

PR #198, opened 2022-06-17
Fix DBLoad result task

Rickmarges:
## Problem
The "DBLoad: Installation results" task fails due to a variable in the when statement not existing. But on inspection the wrong variable is checked in the when clause

## Solution
Replace the variable for the correct one


Rickmarges:
Sorry, I might have been too quick. I may have ran some old code. Will test this further

PR #202, opened 2022-06-20
Bugfixes

Rickmarges:
## Problem
- Suse grub parameters are applied to RHEL systems
- lv_usrsap isn't created on any machine

## Solution
- Add conditional to only apply the Suse grub config when needed
- Update conditionals and list


hdamecharla (approved):
Hi @Rickmarges , thank you for the PR. The proposed changes look good to me, approved 🙏🏾

PR #219, opened 2022-07-19
Kernel parameter kptr_restrict should be set to 1

hjstam:
## Problem
The kernel parameter kptr_restrict is currently set by the code to the value of 2 which is marked as unsafe. I can't find a SAP note that requires this value to be set to 2 so I want to change the value to 1.

## Solution
Set the kernel.kptr_restrict parameter to 1

## Notes
https://sysctl-explorer.net/kernel/kptr_restrict/
https://www.stigviewer.com/stig/red_hat_enterprise_linux_8/2021-12-03/finding/V-230547


hdamecharla (approved):
Hi @hjstam , looks good to me. Approved #219 🙏🏾

PR #222, opened 2022-07-20
Correct name field in S4 hana bom files

bajajyukta7:
Problem
Names are not uniform across bom files.

Solution
Follow a standard pattern from name of S/4 hana bom - "SAP S/4 Hana SPS "


hdamecharla (approved):
@bajajyukta7 thank you for the PR. Looks good to me, approved #222

PR #233, opened 2022-08-04
Update bom.j2 to add endline character after filename

bajajyukta7:
## Problem
Filename and target are coming in same line.

## Solution
Add a new line character after filename

## Tests
Tested by running the playbook with new change
image

## Notes


KimForss (approved):

KimForss:
Thanks @bajajyukta7

PR #235, opened 2022-08-05
Remove spaces from sudoers template

Rickmarges:
## Problem
A syntax error occurs in parsing the sudoers file

## Solution
Removed spaces from the template


KimForss (approved):

Rickmarges:
@KimForss we ran into an issue where removing the spaces resolved the issues. After further testing spaces didn't seem to be the issue. It worked with and without the spaces, so I'll close this PR

PR #263, opened 2022-09-05
Added custom DNS for storage_accounts in system

ALTrijssenaar:
## Problem
When using private end points for storage accounts in combination with a custom DNS terraform is unable to find the storage account.

## Solution
Add a custom DNS
Add a time out to wait for the DNS record to succeed

## Tests
Deployed in an environment with PEP and DNS requirements


deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/storage_accounts.tf
```diff
@@ -31,9 +31,50 @@ resource "azurerm_storage_account" "sapmnt" {
   account_tier                    = "Premium"
   account_replication_type        = "ZRS"
   account_kind                    = "FileStorage"
-  enable_https_traffic_only       = false
+  enable_https_traffic_only       = true
```
KimForss, line 34: For Azure Files NFS this has to be false

https://docs.microsoft.com/en-us/azure/storage/files/storage-troubleshooting-files-nfs#cause-2-secure-transfer-required-is-enabled

KimForss (changes requested):
Thanks, one change requested

KimForss (approved):

PR #264, opened 2022-09-06
ignore_changes = [tags]

ALTrijssenaar:
## Problem
Due to policies various resources are applied with required tags. Even if they change by policy there should be no difference detected by TerraForm.

## Solution
Apply the following code to the resources:
  lifecycle {
    ignore_changes = [tags]
  }
## Tests
Deployed locally


KimForss (approved):
Thanks

PR #269, opened 2022-09-08
Corrected the condition job_result.rc 

ralwani:
## Problem
The step file was not created as condition was job_result.rc > 0  .

## Solution
job_result.rc > 0  ---> job_result.rc == 0

## Tests
Install should create a step file for DBLoad step if its executed successfully

## Notes


KimForss:
thanks @ralwani

PR #288, opened 2022-09-21
🐛 Use bom configuration set by system

UncleBats:
## Problem
The BOM passed through the pipeline options was not used to fill out the sap-parameters.yaml. 
Upon investigation, it made no sense to be able to set this from a pipeline variable cause you want it as part of the system configuration, making it consistent every time you deploy the system and not by mistake choose a different BOM.

## Solution
- Remove from system deployment pipeline
- Remove the default empty variable configuration for bom-name. You want to have a clear message telling you what variable you are missing in the TFVARS when deploying. Not running into issues a deployment later that a bom-name is missing from the sap-parameters.yaml file.
- Add input=false to the terraform apply and plan. To make sure the pipeline is not waiting for manual input.

## Tests
Run pipeline 03, deploy a system that does not exist. If you didn't provide a bom-name it will fail. If you do provide a bom-name that same name should also appear as bom-name in the sap-parameters.yaml that will be used during software installation.

## Notes
Removed your .vscode/settings ;)

### DIFF
--- a/deploy/pipelines/05-DB-and-SAP-installation.yaml
+++ b/deploy/pipelines/05-DB-and-SAP-installation.yaml
@@ -189,12 +189,17 @@ stages:
                 cd $(Build.Repository.LocalPath)/$(Deployment_Configuration_Path)/SYSTEM/$(sap_system_folder)
                 sap_params_updated=0
 
-              echo -e "$green--- Add BOM Base Name and SAP FQDN to sap-parameters.yaml ---$reset"
-                sed -i 's|bom_base_name:.*|bom_base_name:                 '"$(bom)"'|'  sap-parameters.yaml
-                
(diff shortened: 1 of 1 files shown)


KimForss:
The feedback on this is that the BoM does not really have anything to do with Terraform so we should not use Terraform to populate it, this should be a parameter for the 05 pipeline 

You may predeploy a HANA infrastructure but not know the HANA version/Application you want to deploy. So adding it as a parameter to the pipeline is a more solid approach.

ErickSegaar:
Change the pr after disucssion with @KimForss .

can predefined the bom in the tfvars. and overrule it by using a optional freetextbox in the pipeline parameters. Might be replaced by a configstore in the future

PR #294, opened 2022-09-28
🐛Ensure jmespath exists

ErickSegaar:
## Problem
While running from the azdo agents during the ansible installation, there is a dependency upon jmespath to be installed it is a requirement for the json_query that is used on many occasions. Example provided below

```
TASK [roles-os/1.10-networking : 1.10 Networking - Filter out the values for IPAddresses in json format] *** 
fatal: [x01dhdb00l041]: FAILED! => {"msg": "You need to install \"jmespath\" prior to running json_query filter"} 
fatal: [x01scs00l411]: FAILED! => {"msg": "You need to install \"jmespath\" prior to running json_query filter"} 
fatal: [x01app00l411]: FAILED! => {"msg": "You need to install \"jmespath\" prior to running json_query filter"} 
```
## Solution
However jmespath is installed on the agents running the scripting, python is not able to find them in the correct context. By adding them as a 'present' in the current ansible scripting this issue is resolved.

## Tests
run the 05-pipeline on the default public azdo agent all the way until software acquisition

## Notes


deploy/ansible/playbook_02_os_sap_specific_config.yaml
```diff
@@ -12,7 +12,11 @@
     - vars/ansible-input-api.yaml                               # API Input template with defaults
 
   tasks:
-
+    - name: ensure jmespath is installed in the playbook python
+      pip:
```
KimForss, line 16:
```suggestion
      ansible.builtin.pip:
```

KimForss (changes requested):
Can you update the taks to use fully qualified module names

ErickSegaar:
Ok will look intonit tomorrow

KimForss (approved):

PR #297, opened 2022-09-29
Make sure ansible can access azure resources

ErickSegaar:
## Problem
When running the ansible scripts they need access to the resources created in azure. 
When deploying using azure devops there will be an new machine for each stage, they are short lived. Therefor in each stage there should be at least te az login to make sure ansible can do it's job.
Issues with the ansible task

## Solution
Created the variable Use_Deploy_VM on the management plane to be able to test to login using a system managed identity or use the service principle. (when deploy using azdo public/private agents that are shortlived, there is no such thing as a system managed identity available). This variable is being set by deploying the control-plane

Replacing all the ansible tasks with the scripting provides the login into azure and proper call of the ansible playbooks.

## Tests
Run the 05-pipeline with all options checked

## Notes

### DIFF
--- a/deploy/pipelines/01-deploy-control-plane.yaml
+++ b/deploy/pipelines/01-deploy-control-plane.yaml
@@ -376,14 +376,6 @@ stages:
                        file_key_vault=$(cat ${deployer_environment_file_name} | grep keyvault= | awk -F'=' '{print $2}' | xargs)
                        echo 'Deployer Key Vault' ${file_key_vault}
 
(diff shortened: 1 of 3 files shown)


KimForss:
The deployer cannot be identified via the public IP, what we can do is to check if the deployer has been setup via us, if it is it will have the file ' /etc/profile.d/deploy_server.sh' present, we need to change the logic to use that file

ErickSegaar:
Ok will change to look for that file

ErickSegaar:
Reworked this, but need to at least run a full deployment to validate. Replaced copy & paste by pipeline templates, if it itches you should scratch :)

KimForss (approved):

PR #300, opened 2022-09-29
Remove async job wait db_load

ErickSegaar:
## Problem
During the wait of the async db load job, anywhere between 3-15 minutes it would result with the error

```
TASK [roles-sap/5.1-dbload : DBLoad Install -Wait for asynchronous job to end] *** 
task path: /agent/_work/1/s/deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml:150 
fatal: [x01app00le41]: FAILED! => { 
"ansible_job_id": "397969119419.703904", 
"attempts": 12, 
"changed": false, 
"finished": 1, 
"invocation": { 
"module_args": { 
"_async_dir": "/home/azureadm", 
"jid": "397969119419.703904", 
"mode": "status" 
} 
}, 
"msg": "could not find job", 
"results_file": "/home/azureadm/397969119419.703904", 
"started": 1, 
"stderr": "", 
"stderr_lines": [], 
"stdout": "", 
"stdout_lines": [] 
} 

```

## Solution
Removing the async wait resolved the entire issue. Dont know why it was there in the first place case it is a pretty much synchronous operation

## Tests
Run the db_load in pipeline 5

## Notes

### DIFF
--- a/deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
+++ b/deploy/ansible/roles-sap/5.1-dbload/tasks/main.yaml
@@ -144,30 +144,18 @@
       register:                        dbload_results
       tags:
         - skip_ansible_lint
-      async:                           7200
-      poll:                            0
-
-    - name:                            "DBLoad Install -Wait for asynchronous job to end"
-      ansible.builtin.async_status:
-        jid:                           "{{ dbload_results.ansible_job_id }}"
-      register:                        job_result
-      until:                           job_result.finished
-      retries:                         120
-      delay:                           60
-      when:                            dbload_results.ansible_job_id is defined
 
     - name:                            "DBLoad: Installation results"
       ansible.builtin.debug:
         msg:
           - "DBLoad : {{ dbload_results }}"
-          - "Job result: {{ job_result }}"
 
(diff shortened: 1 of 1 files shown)


KimForss:
The async wait is in place as the dbload might run over 120 minutes which would cause the task to fail, with async - there is a "life sign" and the task will not fail

ErickSegaar:
> The async wait is in place as the dbload might run over 120 minutes which would cause the task to fail, with async - there is a "life sign" and the task will not fail

Will look into it again on Monday, there might be an issue with life time of the pipeline aswell then

ErickSegaar:
@KimForss based on your comments, I reran the standalone with the await in the pipeline. Now it ran successful, so I guess my observations of this process failing where due to another configuration issue that I solved down the line and did not track back to find if this was still an issue.

I will close this PR

ErickSegaar:
Error on my side, mixed up the DB installation with the Database_load. Still failed with the async. Will try and figure out where it goes south but issue is still here

ErickSegaar:
Did some more testing, Our hardenend images only had 2GB on root. That is not enough when running the async tasks, it will fill up and crash. After adding 10GB to the root, both deployments worked fine. So I will pull the PR.

PR #301, opened 2022-09-29
make sure folder settings apply

ErickSegaar:
## Problem
During execution the process did not have access to the /usr/sap folders. 

## Solution
Tried to figure out where the original settings where applied to change it but this was quicker, just chmod before execution

## Tests
Run the db-install from the 05 pipeline

## Notes
If you know where these folders are setup in the first place it might me better to fix it there instead of doing it my way.


KimForss:
Is this problem only present for HANA installations?

KimForss (approved):

PR #304, opened 2022-10-05
🐛 ✨ 🎨 📝 Separate config and code repos

ErickSegaar:
## Problem

- Synchronising the sap-automation repository to the client's repository is a pain. It is a tight integration between a framework and configuration 
- Copy-paste pattern and complex variable usage throughout the 05 pipeline for logging and downloads, leading to making changes to multiple places making it hard to read and maintain.
- After introducing the plan properties for vm's this resulted in hard deletes due to a forced destroy when changed
- When deploying a single instance the xfs mount point will be overwritten by an nfs mount point (that will not work on a standalone)
- vm-size for S4Demo was big enough to deploy SAP (needed more memory)
- Small bugs (persisted merge conflicts, removed shared disk configuration, DNS records, etc)

## Solution
- Separate client configuration into a separate repository, by which a client cannot make any change by mistake in the sap-automation repository. Link the client's repository to use the sap-automation repository giving the client the possibility to change versions of sap-framework used based upon git references (branches, tags, commits). A boilerplate for the creation of the client repository is added and the document is provided in the MD files.

Reworking the pipelines <= pipeline 12 to be a template to be consumed by the clients own pipeline and reworking the scripting to differentiate between the sap-automation-repo and the client-config-repo
- Moving download and log collection into their own separate template to be reused. Fine-tuning the discovery of log files and pushing them back as artifacts into the pipeline. Remove the push to git repo of the logfiles cause they are now properly assigned to a deployment in the azuredevops pipeline artifacts

- Adding a new property "type" to the image definition used can be 3 different:
1. custom = use a organisation's own hardened image by providing the source_image_id
2. marketplace = use a marketplace hardened image, this will set the plan property on the vm
3. source_image = use a source_image provided by ms and will set the source_image_reference on the vm
```
application_server_image = {
  type            = "custom" # [custom, marketplace, source_image]
  os_type         = "linux"
  source_image_id = "/subscriptions/xxx-xxx-xxx/resourceGroups/myresourcgroup/providers/Microsoft.Compute/galleries/mysharedgalary/images/customImageDefinitoin/versions/latest"
}
```
- Adding additional condition to not configure the nfs4 mount point when configuring a standalone
- Updating the S4Demo to use a bigger vm-size
- removed the persistence of the .terraform/terraform.tfstate files (only the zips are required for the automation)
- add explicit fail on error for internal scripts, no pipeline should turn green if it is not prestine

## Tests
Running and deploying distributed and stand-alone systems using the azure devops pipelines <= pipeline 12

## Notes
todo before pr completion:
- [x] write markdown for configuration clients repository
- [x] write boilerplate for configuration clients repository
- [ ] ~~remove BOM definition from 05-pipeline parameter~~
- [x] rework pipeline 11 into templated version
- [x] rework pipeline 12 into templated version


ErickSegaar:
We transformed the 11 and 12 but ran into some state issues that we think already existed in the original files. So pls take a look at it after the completion of this pr

KimForss (approved):
Approved

PR #308, opened 2022-10-10
ACSS registration 

mkdeegan:
## Problem
Desire to auto-register SID to ACSS, post SAP install.

## Solution
Extend pipeline "05-DB-and-SAP-installation" with a stage and job to affect the ACSS Discovery process.

## Tests
Perform install of SAP using the pipelines.
A checkbox is now available as part of the Pipeline parameters (Default: True).
Two additional parameters are required;
- A Prod/NonProd selection (Default: NonProd)
- System Type selection: S4HANA|ECC|Other (Default: S4HANA)

## Notes
This uses an Azure CLI Extension to call the ACSS Discovery tool from the pipeline using the agent on the Deployer.

### DIFF
--- a/deploy/pipelines/05-DB-and-SAP-installation.yaml
+++ b/deploy/pipelines/05-DB-and-SAP-installation.yaml
@@ -87,6 +87,28 @@ parameters:
     type:                              boolean
     default:                           false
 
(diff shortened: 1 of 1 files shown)


hdamecharla (approved):
@mkdeegan looks good to me, approved #308 🙏🏾

PR #319, opened 2022-10-28
🚀✨ Support nightly-builds

ErickSegaar:
## Problem
- when running a nightly pipeline, just sequencing all the steps. The pipeline does not continue if a stage is skipped due to default behavior.
- Some lifecycle ignore tags where removed and are forcing redeploys on each run

## Solution
- Adding a nightly.yml to the boilerplate + description
- Adding conditions on each first stage of the pipelines
- re-adding the ignore tags
- Added the parameter to force a template to get the latest version from a branch instead of the sha it started on. (when ever a pipeline starts it will execute all checkouts based on the SHA it started on. When running everything in 1 pipeline for the nightly you need the sap-parameters.yml in the 05 when it was added in the 03. So you need to force a latest pull)

## Tests
Run the nightly

## Notes


ErickSegaar:
Going away present 🎁

KimForss (approved):
Looks good to me

PR #323, opened 2022-11-07
Update HANA_2_00_install.rsp.j2

mimergel:
## Problem
HANA installation fails with errors:

    Master Password cannot be used in combination with other initial passwords

## Solution
remove reference of {{ main_password }} for all other initial passwords and
keep only the Master Password
master_password={{ main_password }}

## Tests
done on SLES 15 SP3 command line and via devops pipeline run

## Notes
reason is unclear why hdblcm changed the behavior


mimergel:
replaced with new PR to experimental

PR #326, opened 2022-11-09
Failure of pipeline 10. Remover-terraform.yaml

soursarkar1990:
## Problem
The pipeline was failing due to wrong local variable in different files

## Solution
Removed the application variable to application_tier

## Tests
Tested with and without the change and it's working

## Notes


KimForss (approved):
Thank you

PR #330, opened 2022-11-11
Fix for installer.sh failing when running "sap_library" stage greenfield deployment .

shekharsorot:
## Problem
When performing a deployment via scripts, sap_library deployment will fail with below message immediately after terraform plan. This happens when no previous sap_library state file exists locally or remotely.

<< EOF >>
╷
│ Warning: No outputs found
│
│ The state file either has no outputs defined, or all the defined outputs are empty. Please define an output in your configuration with the `output` keyword and run `terraform refresh` for it to become
│ available. If you are using interpolation, please verify the interpolated value is not empty. You can use the `terraform console` command to assist.
╵
╷ │ Warning: No outputs found │  │ The state file either has no outputs defined, or all the defined outputs are empty. Please define an output in your configuration with the `output` keyword and run `terraform refresh` for it to become │ available. If you are using interpolation, please verify the interpolated value is not empty. You can use the `terraform console` command to assist. ╵
The subscription of '╷ │ warning: no outputs found │  │ the state file either has no outputs defined, or all the defined outputs are empty. please define an output in your configuration with the `output` keyword and run `terraform refresh` for it to become │ available. if you are using interpolation, please verify the interpolated value is not empty. you can use the `terraform console` command to assist. ╵' doesn't exist in cloud 'AzureCloud'.
Trying to find the storage account
Warning: No outputs found

<<EOF >>

This is due to lack of test condition in file installer.sh : 630.  Script attempts to fetch 'terraform xxxx output'  of a state file that either is not present or populated with actual output data block.
This will cause continuous failure unless Terraform is run manually with correct backend parameters. 
Repeated execution of 'installer.sh' or 'prepare_region.sh' will continue to throw same error with no recovery possible without manual intervention.

## Solution
Add a condition to the block in installer.sh: 630 to 642. Use the variable $deployment_parameter to check if this is a new deployment of not. In case of new deployment, do not execute this entire block. 
$deployment_parameter is already being populated to identify if a given bootstrap is fresh deployment or existing.

## Tests
Can successfully deploy sap_library  on a fresh environment. Existing sap_library state file when present is being correctly looked up from Azure storage account.

## Notes

Add fixes to various IF conditions, where due to usage of single square brackets is causing Bash to throw errors about expecting unary operator. This has been fixed by enclosing variables within the brackets in double quotations. This ensures POSIX standard compliance along with Bash on WSL as well as Ubuntu VM

example : IF [ $variable == "test" ]   changed to IF [ "$variable" == "test" ] where ever script throws error during execution.


hdamecharla:
Hi @shekharsorot , thanks for the PR. Could you target the experimental branch with this PR, please?

PR #337, opened 2022-11-14
Connection_name is not a variable any more

soursarkar1990:
## Problem
The 01. Deploy control plane was failing

## Solution
Connection_name is now a parameter and not a variable.
use_deployer is no longer required as it's been removed from the parameter section

## Tests
Tested it with and without the changes.

## Notes


hdamecharla (approved):
looks good to me, approved.

PR #339, opened 2022-11-15
Experimental Branch : landscape subnets.tf issue where deployment will fail when using existing subnets in brownfield scenario.

shekharsorot:
## Problem
In a brownfield deployment when using existing subnets, TF code will fail with error on "azurerm_route_table_association" blocks as the route tables are not created in the "azurerm_route_table" blocks given success of the test conditions.

## Solution
replicated the conditional checks in azurerm_subnet_route_table_association  admin/db/app/web blocks from "azurerm_route_table" block

## Tests
run the Landscape deployment with existing subnets already in place. Original code should fail with error ""azurerm_route_table.rt[0].id does not have an entry with specified index"
Adding checks allows us to use existing subnets ( leaving them untouched ) while creating and attaching route table table to subnets created by Terraform.

## Notes


KimForss (approved):
Looks good to me

KimForss:
Thanks

PR #346, opened 2022-11-17
Experimental : SAP System deployment fails when specifying an existing keyvault in brownfield deployment.

shekharsorot:
## Problem
When deploying sap system and specifying an existing keyvault ( for brownfield deployment ), Terraform code attempts to create the keyvault and fails due to keyvault name already existing in Azure

## Solution
Add conditional checks for keyvault creation and output for exporting keyvault ID

## Tests
Execute the deployment with a keyvault that exists // maintained outside terraform

## Notes


KimForss (approved):
Looks good to me

PR #348, opened 2022-11-22
Variable name change

soursarkar1990:
## Problem
Deploy control, SAP system deployment and removal pipeline was failing due to variable change

## Solution
The variable name was changed as CP was added in front of them

## Tests
tested with and without the new variables

## Notes


KimForss:
The CP_ prefix was added to distinguish between the control plane SPN and the Workload zone SPN. 

The SDAF-MGMT variable group (or whichever group is hosting the control plane will have the variables as CP_ and the workload variable groups will have ARM_CLIENT_ID etc.

PR #359, opened 2022-12-05
Fix linting and require ansible-lint 6.9.1

hjstam:
## Problem
Ansible Lint was failing

## Solution
Fix linting violations


KimForss (approved):
Thanks :)

PR #360, opened 2022-12-06
Adding BOM for Netweaver 750 SPS25

shekharsorot:
## Problem
Adding BOM for Netweaver 750 SPS 25 based on recent customer ask & non-prod deployment

## Solution
BOM created based on guidelines set in "https://learn.microsoft.com/en-us/azure/virtual-machines/workloads/sap/automation-bom-get-files"

## Tests
Deployment tested in a customer non-prod environment currently in use.

## Notes
Uses existing HANA_2_00_065_v0002ms, SUM20SP15_latest and SWPM10SP36_latest (version 3)  BOMs


hdamecharla (approved):
Checksums need to be added but looks good to me. Approved.

PR #362, opened 2022-12-06
Fix enabling/disabling of maintenance mode for scsers/hanadb clustering

hjstam:
## Problem
For Red Hat the cluster was only taken out of the maintenance mode but it was never in maintenance.

## Solution
Set the cluster in maintenance first.

## Tests

## Notes


KimForss (approved):
Thanks

PR #378, opened 2022-12-23
Added the changes for fixing DB Discovery issue in HA scenarios.

ShayakSarkar:
## Problem
The sap profile parameters were not being set correctly after the DB Load installation. The SAPDBHOST param in particular was being set to the primary DB VM instead of it being the DB LB. This was preventing all the DB VMs backing a DB instance from being discovered. The SAPDBHOST is set to the primary DB VM so as to ensure reliable installation of the DB Load. After the installation succeeds, the SAPDBHOST value should be set to point to the DB LB. This was not being done.

## Solution
A new file 4.0.1.7-sap-profile-changes.yml is added which contains tasks for setting the sap profile parameters correctly after the DB Load installation. These tasks are executed at the beginning of PAS install.

## Tests
Manually tested and E2Es refined in the service code to test whether all DB VMs are being discovered in the post installation discovery.

### DIFF
--- a/deploy/ansible/roles-db/4.0.1-hdb-hsr/tasks/4.0.1.7-sap-profile-changes.yml
+++ b/deploy/ansible/roles-db/4.0.1-hdb-hsr/tasks/4.0.1.7-sap-profile-changes.yml
@@ -1,0 +1,66 @@
+---
+
+##############################################################################################################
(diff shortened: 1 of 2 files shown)


hdamecharla (approved):
As these changes were pulled in from the main branch, approving the PR #378

PR #380, opened 2022-12-26
Fix BOM target file name parameter from 'override_target_filename' to 'filename'

shekharsorot:
## Problem
When processing BOM for SWPM windows, ansible skips rename of downloaded file to SWPM.SAR as BOM contains 'override_target_filename' property which is not used by "roles-sap/windows/3.3-bom-processing"

## Solution
replaced property 'override_target_filename' with 'filename'

## Tests
run play_bom_downloader , then playbook_03_bom_processing to validate that the archive is unpacked correctly.

## Notes
minor fix to BOM


hdamecharla (approved):
@shekharsorot #380 Looks good to me, approved.

PR #383, opened 2022-12-29
Experimental : update 0.1-bom-validator /tasks/bom_download.yaml to correctly create checksums for custom BOMs

shekharsorot:
## Problem
When downloading a custom BOM with "create_checksums" option, playbook will fail with error as mentioned in section Additional notes

## Solution
Since regexp is now deprecated as per Ansible documentation (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/blockinfile_module.html ), the line bom_download.yaml:123 is not longer valid. the property "insertafter" can use regular expression to locate where to move the file cursor to.
Line 124 has been changed from " insertafter:               '      archive:      {{ item.archive }}'"  to "insertafter:               '^\s*archive:\s*{{ item.archive }}'"
Line 123 is commented out.

## Notes
Ansible documentation link:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/blockinfile_module.html
Error when using older Ansible API

################ Error #####################
atal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ansible.builtin.blockinfile) module: regexp. Supported parameters include: seuser, owner, create, attributes (attr), state, path (dest, destfile, name), insertbefore, block (content), setype, backup, mode, marker_end, marker, group, validate, selevel, marker_begin, insertafter, unsafe_writes, serole."}


hdamecharla (approved):
@shekharsorot PR #383 looks good to me. Approved.

PR #386, opened 2023-01-03
Syntax errors fixed in JAVA stack templates ( ers and app) for NW750_SPS25

shekharsorot:
## Problem
malformed templates caused errors when deploying app and ers in distributed setups.

## Solution
fixed the syntax errors in the param templates.


hdamecharla (approved):
Approved.

PR #390, opened 2023-01-09
Adding BOM for Erp6 ehp8 linux for non-HA/Prod setup

shekharsorot:
## Problem
Customer requested a BOM for ERP6 based on HANA 2_0 DB. Uploading BOM as current ERP6 BOM's are based on MSSQL and/or DB2 and missing HANA DB based deployment.

## Solution
BOM Build designed from SAP maintenance planner and templates tested on Single and distributed non-HA setup


hdamecharla (approved):
Approved

PR #397, opened 2023-01-16
Nw750 java fixes

shekharsorot:
## Problem
Fix various issues in the BOM's templates files. Added variable in BOM Yaml to signify that this is a JAVA deployment.
Fixes are brought in directly from a customer deployment.


KimForss (approved):
Changed the target to experimental

PR #399, opened 2023-01-17
Update playbook_00_validate_parameters.yaml

nnoaman:
Ensure excluding the Windows machines

## Problem
Windows target machines are getting an error from this Ansible task, since it should be executed only in the Linux related machines.

## Solution
Adding a condition to exclude the Windows machines.


hdamecharla (approved):
Hi @nnoaman , PR#399 looks good to me. Approved.

PR #401, opened 2023-01-19
Use bom_processing_become variable and fix microsoft_supplied_bom_archive path variable

hjstam:
## Problem
- BOM download isn't using the bom_processing_become  everywhere
- microsoft_supplied_bom_archive.path isn't a valid variable

## Solution
- Use bom_processing_become variable
- Fix microsoft_supplied_bom_archive variable


hdamecharla (approved):
Approved #401

PR #402, opened 2023-01-20
SAP software provisioning manager (sapinst) should be run with umask 022

hjstam:
## Problem
SAP software provision manager (sapinst) should be run with umask 022 when executing as the root user according to the documentation (see notes).

## Solution
Add umask command before every sapinst command with the possibility to override the default umask via the custom_umask  variable.

## Notes
- https://help.sap.com/docs/SLTOOLSET/06f7611290ea4b8d9bc431a8d7bf05c3/1d4bd354b65ced05e10000000a4450e5.html?version=CURRENT_VERSION_SWPM20


hdamecharla (approved):
Looks good, approved #402

PR #408, opened 2023-02-01
Cluster resources control & Terraform formatting

KimForss:
## Problem
Sometimes the SCS cluster resources are on the secondary node and the MessageServer is unable to start on the primary node.

## Solution
This PR adds a check on the SCS cluster status and it will move the resources if needed.

## Tests

## Notes


hdamecharla (approved):
Approved #408

PR #410, opened 2023-02-06
NW on Windows Enhancement

nnoaman:
## Problem
An error related to install a missing Visual C++ redistributable package occurs after running sapinst. The error terminates the SAP installation, since it requires a manual reboot first and running the sapinst again.

## Solution
Ensure installing the packages and performing an auto reboot before running sapinst.

## Tests
The solution tested successfully.


hdamecharla (approved):
@nnoaman #410 looks good to me. Approved 🙏🏾

PR #423, opened 2023-02-24
Update to SUSE Hana - HA cluster creation : force primary node to be master before completing task set.

shekharsorot:
## Problem
In SLES based HA - HSR ,  the SAPhanaSR script hook (from SUSE)  under some cases will make secondary node as primary during cluster creation.  This causes DBLoad playbook to fail as the task_3.3.1 (BOM Utility) populates the DB name as primary VM instead of Cluster name.
This is not being seen in RedHat based HSR clustering ( using PCS ).

## Solution
The fix is to force HANA to move to master using the below Jinja2 shell sequence : 
"crm resource move msl_SAPHana_{{ db_sid | upper }}_HDB{{ db_instance_number }} {{ primary_instance_name }} force"
right before removing the cluster from maintenance mode. The resource constrains + automatic_failover: false should prevent any automatic movement of HANA master role to secondary VM unless requested by cluster admin.

Alternate workaround ( not implemented at the moment ) is to add logic to taskset 3.3.1 ( BOM  utility) to populate SAP HA virtual name instead of HANA primary VM name in the DBload template.

Another workaround is to do a DB load before HSR creation , this could however increase cluster creation time as DB backup and restore will take long time. The effect of this approach is not known on Scale out setups ( NetApp or HA+HSR ) approach and needs to be investigated.

## Tests
Previously, 2 in 5 HANA deployments will cause secondary node to become master during HSR clustering. Currently tested infra showed subsequent deployments forcing primary node to be actual HANA master. The behavior is more prevalent in VM SKU where HANA takes a long time to start post the Stop/Start  sub tasks.

## Notes
Recommend another round of testing before merging into main.  Although this command should not cause any issues or incompatibilities with current working scenarios as there is no actual DB replication going on at that stage and role migration to node if applicable is instantaneous. If Primary node is Master, this command is ignored.

Its important that DB load or any data import activity is done post cluster creation. If HSR - HA is being done on HANA with already existing data, ensure that the database is already replicated to the second Hana before running HSR/HA setup.

RHEL will use "pcs" instead of "crm", currently attempting to replicate problem in RHEL environment.

This change is referencing the official SLES HA guide at link : "https://learn.microsoft.com/en-us/azure/sap/workloads/sap-hana-high-availability'

Update: This minor fix is currently being used in a customer deployment


hdamecharla:
Hi @shekharsorot , I think this is more of change in the order of execution. I will get back to you on this. Until then, can you put this PR in draft state please?

cc: @nnoaman , @KimForss , @mkdeegan

shekharsorot:
putting PR in draft.

shekharsorot:
removing PR , patch to be pushed with equivalent fix for RHEL HANA cluster added to SLES

PR #424, opened 2023-02-28
Bugfix : HSR setup failing when OS name is in upper case

shekharsorot:
## Problem
When HANA VM is provisioned with OS name in upper case, HSR setup will fail with messages "SAPSYSTEMNAME not in profile".
On detailed examination, users are able to run "hdbnsutil" from shell running as <sid>adm but not when triggered from ansible.

Reason : Irrespective of OS name, HANA setup will create and populate the profile directory under /usr/sap/<SID>/HDB<instanceID>/ as lower case OS name. More so, sapprofile.ini under the same folder has SAPSYSTEMNAME also populated in lower case. 
When running commands via Ansible task : 4.0.1.5, the environment variables are generated using Ansible variables that denote actual hostnames coming from <SID>_hosts,yaml which are in upper case.
This results in "hdbnsutil" executions to fail as it points to a directory that does not exists and does not have the sapprofile.ini file in it with correct params.

## Solution
In file "4.0.1.5-provision_hana_replication.yaml", ensure SAP HANA profile directories are correctly populated by correcting the case of hostname , denoted by {{ ansible_hostname }}

## Tests
Provision VM with upper case hostnames matching the same upper case OS Names. HANA HSR setup will fail.
Apply fix and HANA HSR setup will proceed without errors.

## Notes
SAP documentation link for paths to profile directory and configuration file ( https://help.sap.com/docs/SAP_HANA_PLATFORM/eb3777d5495d46c5b2fa773206bbfb46/aa7e697ccf214852a283a75126c34370.html?version=1.0.12)

### DIFF
--- a/deploy/ansible/roles-db/4.0.1-hdb-hsr/tasks/4.0.1.5-provision_hana_replication.yml
+++ b/deploy/ansible/roles-db/4.0.1-hdb-hsr/tasks/4.0.1.5-provision_hana_replication.yml
@@ -29,14 +29,14 @@
     HOME:                              "/usr/sap/{{ db_sid | upper }}/home"
     PYTHONHOME:                        "/usr/sap/{{ DB }}/exe/Python3"
     DIR_EXECUTABLE:                    "/usr/sap/{{ DB }}/exe"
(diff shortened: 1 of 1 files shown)


nnoaman:
Hi Shekhar, LGTM. Thanks!

PR #425, opened 2023-03-01
Add support for DB2 deployments on RHEL8.7

hjstam:
## Problems
- No support for Red Hat Enterprise Linux 8.7 (which is supported by SAP / DB2 deployements)
- BOM processing fails when using a deployer without root permissions (again)
- Support deployer without /mnt disk
- OS configuration done flag tasks should run without root permissions
- 2.6.0-afs-mounts role is included for every other NFS type as ANF, should be AFS only
- BOM downloader fails when using a deployer without root permissions
- DB2 installation fails when /tmp has noexec mount option
- Ansible-Lint fails
- Github action workflow fails

## Solution
- Add Red Hat Enterprise Linux 8.7 (RHEL 8.7) as operating system
- Add bom_processing_become variable to allow bom processing without root permissions
- Add when statement to check free diskspace of /mnt filesystem
- Remove become parameter for OS configuration done flag task
- 2.6.0-afs-mounts role is included only when NFS type is set to AFS
- Add bom_processing_become variable to BOM downloader
- Add DB2TMPDIR to set temporary db2 installation location
- Remove skip_ansible_lint tags and add noqa comments when necessary
- Remove requirements.yml from workflow directory, requirements.yml in ansible directory is enough

## Tests
- Deploy a DB2 stack on RHEL 8.7 (custom azure image) with /tmp mounted as noexec. NFS option set to NFS (custom)

## Notes
- Ansible-Lint actions still shows warnings like 'Unable to load module ansible.windows.win_find at deploy/ansible/roles-sap/windows/5.1-dbload/tasks/main.yaml:265 for options validation'. I believe this has something to do with ansible-lint v 6.11 and should be fixed in v6.13 but there is no Ansible Lint for GitHub Action available for this version yet.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -3,8 +3,6 @@ exclude_paths:
   - deploy/ansible/BOM-catalog
   - deploy/terraform
   - .github
-  # added ansible_skip_lint tag to exclude the task rather than the whole file.
-  # - deploy/ansible/roles-sap-os/2.10-sap-notes/tasks/2.10.1.1.yaml # Has to be excluded as there is a pipe sign in the regex that Ansible sees as filter
 
 warn_list:
   # FIXME: Experimental violations should be fixed in separate PR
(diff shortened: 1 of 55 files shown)


hdamecharla (approved):
#425 - approved

PR #427, opened 2023-03-02
Directory permissions sap_deployment_automation should be set to 0755

hjstam:
0766 is useless as the other user has no execute permissions and can't traverse into the directory. On all other places in the code 0755 is specified, lets keep it the same.


hdamecharla (approved):
LGTM #427 approved.

PR #435, opened 2023-03-28
Fix hardcoded HANA port in hdbuserstore list

shekharsorot:
HANA port in hdbuserstore is hardcoded to 30313 for App and PAS install tasks. Fixed to correctly generate correct port for deployed HANA instances.

## Problem
Issue reported by a customer. Customer notified us that hdbuserstore has incorrect HANA port assigned to cluster name.
Investigation revealed static port assignment of 30313 which assumes HANA instance ID is always 03.

## Solution
Fixed by assigning logic to generate port using ansible vaiable "db_instance_number"


hdamecharla (approved):
LGTM, approved. Thank you, @shekharsorot

PR #465, opened 2023-06-30
Provide several fixes for Terraform and Ansible

Rickmarges:
## Problem
1. Terraform contained several problems when not using a deployer as well as some outdated version pins
2. The check for internet connection was always executed
3. When downloading the BOM it would be set to the `ansible_orchestrator_user` which isn't needed when running without become
4. `__pycache__` was included for commits
5. An unnecessary become was used on a check
6. The DNS loadbalancer output wasn't containing the SCS loadbalancer when not deploying the web
7. There is a become on a block of the installation of the APP, but not in the right place
8. A symlink is created for `compact-sap-c++-10`, but the owner and group was set on the source instead of the symlink
9. A `sapmnt` mountpoint is created on HANA systems which isn't needed

## Solution
1. Add/fix try statements and update version pins
2. Removed a duplicate task and added `check_outbound` to the when statement
3. Simplified some loops, removed duplicate tasks, and only set owner to `ansible_orchestrator_user` when become is true
4. `__pycache__` is added to the gitignore
5. Set the become to false
6. Fixed the zipmap/concat/slicing of the output variables
7. Moved the become to the correct position
8. Set `follow` to false to set the owner and group of the symlink itself
9. Updated the when statement to also exclude HANA

## Tests
We've deplyed a single and HA HANA stack without problems

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,8 @@ Webapp/SDAF/Properties/PublishProfiles
 Webapp/.vs/ProjectEvaluation
 Webapp/SDAF/Properties/ServiceDependencies
 
-
 # Ignore OS files         - MKD 20221221
 **/.DS_Store
+
+# Ignore Python cache directories
+__pycache__
--- a/deploy/ansible/playbook_00_validate_parameters.yaml
+++ b/deploy/ansible/playbook_00_validate_parameters.yaml
@@ -408,26 +408,6 @@
 
     - name:                            "0.0 Validations - Check internet connectivity"
       ansible.builtin.uri:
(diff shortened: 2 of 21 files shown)


hdamecharla:
@Rickmarges thank you for the PR. For larger PRs we would like to test these before pushing to main so, could you target experimental branch for this please?

Rickmarges:
@hdamecharla I've changed the target branch to experimental

hdamecharla (approved):
@Rickmarges Thank you, for the PR. Approved #465 for merge into experimental.

PR #466, opened 2023-07-03
Fixing Logical Volume creation order

jaskisin:
## Problem
logical volumes gets created in wrong order. usrsapinstall lv gets created first of all the lvs for vg_sap due to which usrsapinstall gets all the 100% free space from vg_sap

## Solution
fixed the order of the logical volume in such as way that usrsapinstall always comes at last and gets 100% of the free space left after creating logical volumes for sapmnt and usrsap

## Tests
End 2 End testing was performed on Single, distributed and High Availability senarios for RHEL as well as SUSE

## Notes


hdamecharla (approved):
@jaskisin #466 LGTM approved

PR #471, opened 2023-07-17
removing unnecessary characters from japan region codes

jhajduk-microsoft:
## Problem
bug 470 - unnecessary characters in japan region codes

## Solution
characters were removed


hdamecharla (approved):
approved

PR #474, opened 2023-08-07
Several fixes while testing with Ansible 2.14.6

hjstam:
## Problem
- The warn args variable isn't supported in newer Ansible versions
- Grub file lookup was executed on Ansible controller (deployer) not the Ansible host
- Regex search in BOM aggregation contains a invalid escaping of the forward slash character
- Azure metadata is fetched with curl command, this should be handled by the Ansible builtin uri module
- Terraform anydb dns_info_vms output was returning only the ANYDB_SECONDARY_DNSNAME when deploying a single nic VM
- Repo contained a weird 'how' file containing commit information

## Solution
- Remove warn args variable
- Use the slurp module to read the grub configuration on the Ansible host
- Update regex search bij removing escaping of forward slash
- Refactor azure_metadata curl commands to ansible.builtin.uri module
- Refactor anydb dns_info_vms output in the same way as the hana dns_info_vms output
- Remove the how file

## Tests
I tested this PR with a single stack S4HANA deployment with Ansible version 2.14.6.


hdamecharla (approved):
Looks good to me. Approved.

PR #480, opened 2023-08-15
Rsponholtz patch 1

rsponholtz:
## Problem
When retrieving SSH keys from keyvault, the end of line character at the end was removed, resulting in an invalid key format.  It appears that this is true for Ansible 2.8+.  If earlier versions of ansible are still supported, this may cause an issue.

## Solution
Ansible 2.8 adds the stdin_add_newline and strip_empty_ends to the ansible.builtin.command module: https://docs.ansible.com/ansible/devel/collections/ansible/builtin/command_module.html

please set stdin_add_newline to be true, since you can have newlines at the end of a private key, and set
strip_empty_ends to be false.  

while stdin_add_newline should default to true, and strip_empty_ends also defaults to true according to the doc, the command will actially strip the newline from the end if neither of those parameters are set.

## Tests
create a workspace dir, create the sap-parameters.yaml in that with the following content:

```
sap_fqdn: sap.contoso.net
kv_name: rheeee-9e91-2ea147f176e5
secret_prefix: RHE
orchestration_ansible_user: root
sap_sid: RHE
``` 

use the kv that you're actually using.  log onto azure CLI with 

```bash
az login
```

run the ansible using

```bash
ansible-playbook -e _workspace_directory=/root/ansible-test/workspace-dir ./pb_get-sshkey.yaml
```

## Notes
since these options were added in ansible 2.8, if earlier versions are supported it may be nessesary to check the ansible version and use/not use the options


hdamecharla:
Hi @rsponholtz , thank you for the PR. Could you target the experimental branch for the PR please.

hdamecharla:
These changes should be included in SDAF - Release 3.9.0

PR #484, opened 2023-09-11
Set remote user to orchestration_ansible_user

daradicscsaba:
## Problem
The corrected task tries to log in to managed node as root, when in fact it should be the orchestration_ansible_user

## Solution
remote_user value switched from root to orchestration_ansible_user


hdamecharla:
Thank you, for the PR. These changes are already included in SDAF - Release 3.9.0

PR #485, opened 2023-09-20
Fix typo for Japan region deployment

sasukeh:
## Problem
Typo for Japan region at 03-sap-system-deployment.

## Solution
fix typo.

## Tests
Don't need.

## Notes
I expect quick fix this due to timeline for customer.


hdamecharla:
These changes should be included in SDAF - Release 3.9.0

PR #496, opened 2023-10-19
Several improvements and additions for deploying DB2

hjstam:
## Problem
- Set limits for DB2: group name is wrong in hard nproc unlimited
- DB2 installation: is not using the virtual_host defined in the inventory
- DB2 mount: log_dir logical volume is mounted on /log/SID/dir
- Deployer: running the SAP automation without the deployer VM is broken
- enable_firewall_for_keyvaults_and_storage was removed
- VMAgentActions fail
- 2.10-sap-notes: can be optimized and uses invalid oracle8 OS
- SAP OS USERS and Group Creation (nw_user) isn't efficient
- ansible-input-api.yaml has unused variables
- You cannot specify custom virtual_hostnames when deploying a cluster or want to use cluster names when deploying a single stack
- ASCS/ERS NFS filesystem devices are set-up for AFS or ANF use with a specific device path

## Solution
-  Set limits for DB2: use right group name
- DB2 installation: use virtual_host variables
- DB2 mount: log_dir is mounted on /db2/SID/log_dir
- Deployer: add try functions around deployer vm output variables
- Reintroduce enable_firewall_for_keyvaults_and_storage variable
- VMAgentActions don't require become_user and become as they are delegated
- 2.10-sap-notes: optimized and fixed invalid OS
- SAP OS USERS and Group Creation (nw_user) optimized
- ansible-input-api.yaml removed unused variables
- Add custom virtual_hostname variabels for db, ers, scs and pas
- Add variables for ASCS/ERS NFS filesystem devices and add when statements in to restrict set_fact for AFS or ANF

## Tests
Tested with multiple single stack DB2 deployments, single stack HANA and high available HANA stack and it works as expected.

## Notes
Please incorporate this asap in a release. Currently we are also testing the DB2 HA deployments which I would like to do with release.


hjstam:
@KimForss can you take a look at this PR?

hdamecharla:
@hjstam, initial testing of the code reveals HA deployments fail when the custom_[scs,ers,db,pas]_virtual_hostname are not set. This is the default scenario when the code is used without setting these values. The variable ends up having an omit value instead of actually being omitted. The value passed in gets set as '__omit_place_holder__6327d38a8a70754aab5787673a656d0fe2148d47.test.contoso.net'

hjstam:
> @hjstam, initial testing of the code reveals HA deployments for existing scenarios (for example, HANA) fail when the custom_[scs,ers,db,pas]_virtual_hostname are not set. This is the default scenario when the code is used without setting these values. The variable ends up having an omit value instead of actually being omitted. The value passed in gets set as '__omit_place_holder__6327d38a8a70754aab5787673a656d0fe2148d47.test.contoso.net'
> 
> I have created a branch `pull/496` which can be used for testing.

I will try to verify this next week. It might be related to the Ansible version that the SAP Automation is using (2.13) which is EOL by the way. I'm developing and deploying with 2.15.

hjstam:
> I will try to verify this next week. It might be related to the Ansible version that the SAP Automation is using (2.13) which is almost EOL by the way. I'm developing and deploying with 2.15.

@hdamecharla I was able to reproduce the problem and omitting the `scs_virtual_hostname` and `ers_virtual_hostname` variables indeed doesn't work. I was expecting that the role defaults would take over but that isn't the case. 

I've amended the commit and verified that the logic works for the HA SCS/ERS install.

hjstam:
> Hi @hjstam We have updated the codebase with some other changes. You will see that a large number of the terraform files have changed. Both main and experimental have those changes.
> 
> The main change in them is cosmetic so hopefully it does not cause any major issues. I also tried to make sure that our code will work without an deployer.

Thank you Kimmo, I rebased my branch and updated my PR.

KimForss (approved):
Looks good to me

PR #499, opened 2023-10-25
PR for Release v 3.9.3.0

### LINKED ISSUE #494: [BUG] - Bootstrap Control Plane - Key Vault Private Endpoint DNS Record - unexpected status 404 with error: ResourceGroupNotFound
**Describe the bug**
When the control plane is being bootstrapped with shell scripts, the following error is encountered:

```
Error: creating/updating Record Type (Subscription: "<deployer-subscription-id>"
│ Resource Group Name: "<management_dns_resourcegroup_name>"
│ Private Dns Zone Name: "privatelink.vaultcore.azure.net"
│ Record Type: "A"
│ Relative Record Set Name: "<dns-name>"): unexpected status 404 with error: ResourceGroupNotFound: Resource group '<management_dns_resourcegroup_name>' could not be found.
│
│   with module.sap_library.azurerm_private_dns_a_record.kv_user[0],
│   on ../../terraform-units/modules/sap_library/key_vault.tf line 53, in resource "azurerm_private_dns_a_record" "kv_user":
│   53: resource "azurerm_private_dns_a_record" "kv_user"
```

The Subscription ID appears to be incorrect as this should be the dnsmanagement subscription ID specified with the "management_dns_subscription_id" variable. 

I've looked through the code base and it appears that the provider specified in the "azurerm_private_dns_a_record" "kv_user" resource does point to the azurerm.deployer provider. I doubt that this is correct because the private endpoints for storage accounts get created in the correct subscription and resource group (entries specified by  management_dns_resourcegroup_name & management_dns_subscription_id)
**File: sap-automation/deploy/terraform/terraform-units/modules/sap_library/key_vault.tf**

**Expected behavior**
DNS record gets created in the correct subscription

**Additional context**

### LINKED ISSUE #495: Support Custom Tagging for every deployed resource that supports tagging and not just the Resource Group.
**Is your feature request related to a problem? Please describe.**
It would be nice to have a variable that enables custom tagging across all deployed resources, whether it be from the Control Plane, Workload Zone or System. 

**Describe the solution you'd like**
Looking through the code base, I do see tags (via a variable var.infrastructure.tags) only on the ResourceGroup resource. Can this variable be extended to every other resources that supports it, e.g. virtual networks, keyvaults, private endpoints, etc. and then documented?

**Describe alternatives you've considered**
It would be unnecessary to have a separate variable for each resource type, so one variable will fulfill this requirement. For an organization that has strict tagging requirements for resources, this will go a long way to augment resource organization when SDAF is used.

hdamecharla:
## Problem
- No tags being added for all resources.
- Terraform v1.6 is not backwards compatible with Terraform v1.5.6 and reading the statefiles generated by the newer versions break if you are on v1.5.6 or older.
- Due to the way the IP addresses are assigned reading the IP addresses becomes difficult in Ansible.
- Need to assign pipeline permissions for variable groups and agent pools as an improved security measure.
- In some cases, the deployer extension displays the configuration details in the log.
- For command line deployments we do not save the tfvars files centrally. This is not an issue with devops based deployments as they are pushed to the repo.

## Solution
- Tags for all resources.
- Address breaking changes with Terraform v 1.6.
- Fix the secondary IP gathering logic by fetching the information from IMDS.
- Assign pipeline permissions for variable groups and agent pools.
- Fix the issues with the deployer extension.
- Copy the tfvars files into the terraform state storage account.
- General deployment and code readability enhancements for Ansible and Terraform.

## Tests
Tested via pipelines.

## Notes
If you have issues upgrading the control plane VM, remove the configure_deployer extension from the VM manually and redeploy.

## Closes
closes #494 
closes #495


KimForss (approved):
Looks good to me

PR #504, opened 2023-10-31
Set DB2 sapdata stripesize to 32 KB

hjstam:
## Problem
When using LVM striping along with Db2 striping, the extent size of the table space and the stripe size
of the disk should be identical (https://www.ibm.com/docs/en/db2/11.5?topic=management-database-managed-space).
The Page Size is set by default to 16 KB and the Extent size is 2 pages = 32 KB.

## Solution
Set the LVM stripesize to 32 KB.

### DIFF
--- a/deploy/ansible/vars/disks_config.yml
+++ b/deploy/ansible/vars/disks_config.yml
@@ -219,7 +219,7 @@ logical_volumes:
     vg:         'vg_db2_sapdata'
     lv:         'lv_db2_sapdata1'
     size:       '25%VG'
-    stripesize: 256
+    stripesize: 32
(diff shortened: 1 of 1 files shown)


hjstam:
Putting this in WIP as well, we have some unclarities about the strip size vs LVM stripe size

KimForss:
@hjstam This is now configurable, see [Custom Stripe Sizes](https://learn.microsoft.com/en-us/azure/sap/automation/extensibility#custom-stripe-sizes-linux)  Can we close this?

hdamecharla:
Closing the PR as this is now a configurable parameter.

PR #505, opened 2023-11-16
Bug Fix - ensure sapmtn_volume_size is used

SteffenBoThomsen:
## Problem
When using NFS_provider = "NONE" and using NFS exports from the SCS the sapmnt_volume_size is not used for the logical volume.
Example using a `sap` disk of 256g, `usrsap_volume_size = 128g`  and `sapmnt_volume_size = 128g`, the lv gets created with lv_usrsap = 128g and lv_sapmnt = 32g.

## Solution
Change the 
 logical_volumes_sapmnt to conditionally check on `sapmnt_volume_size` and use that if defined else default to the 32g.


hdamecharla (approved):
approved.

PR #508, opened 2023-11-21
Use usrsap_volume_size when specifying NFS_provider NONE

SteffenBoThomsen:
## Problem
When provisioning a sytem using `NFS_provider = "NONE"` and specifying `sapmnt_volume_size` and `usrsap_volume_size` to define the split of `vg_sap`. `sapmnt_volume_size` is used for both logical volumes.

## Solution
Ensure we use the `usrsap_volume_size` for sizing.

A side node on this usage pattern is, that specifying the size in XG eg 128g will fail, as the lv assumes enough Current LE is present - but due to rounding from the vg the second volume group will fail  `Volume group "vg_sap" has insufficient free space (32767 extents): 32768 required` - if using this way of defining the volume group split. It should be set to `sapmnt_volume_size = "50%FREE"` and `usrsap_volume_size = "100%FREE"`. This also ensures that the sap disk on potential databases gets sized correctly.

### DIFF
--- a/deploy/ansible/vars/disks_config.yml
+++ b/deploy/ansible/vars/disks_config.yml
@@ -89,7 +89,7 @@ logical_volumes:
     node_tier:  'all'
     vg:         'vg_sap'
     lv:         'lv_usrsap'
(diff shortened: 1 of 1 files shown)


hdamecharla (approved):
Approved

PR #516, opened 2023-12-15
Bugfix kv-secrets debug and import tasks

hjstam:
Because the operation variable has no default within the rol its existence should be checked.

## Problem
Running the validation playbook errors out:

```
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'operation == \"fencing\"' failed. The error was: error while evaluating conditional (operation == \"fencing\"): 'operation' is undefined. 'operation' is undefined\n\nThe error appears to be in '/opt/sap-automation/deploy/ansible/roles-misc/0.2-kv-secrets/tasks/main.yaml': line 24, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# -------------------------------------+---------------------------------------8\n- name:                                   \"Cluster aware code...\"\n  ^ here\n"}
```

## Solution
Check if the `operation` variable is defined

### DIFF
--- a/deploy/ansible/roles-misc/0.2-kv-secrets/tasks/main.yaml
+++ b/deploy/ansible/roles-misc/0.2-kv-secrets/tasks/main.yaml
@@ -32,7 +32,9 @@
                                             - "use_msi_for_clusters:            {{ use_msi_for_clusters }}"
                                             - "platform:                        {{ platform | upper }}"
     verbosity:                              2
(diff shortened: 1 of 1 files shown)


hjstam:
@KimForss @hdamecharla please add this to main asap, thank you.

KimForss (approved):

PR #523, opened 2024-01-11
Add support for controlling DNS A record creation for secondary names

KimForss:
## Problem
Currently there is no easy way to block the creation of DNS records for the secondary IP addresses
## Solution
Introduce a new variable for Terraform

dns_a_records_for_secondary_names

## Tests
Tested using the DevOps pipelines

## Notes


hdamecharla (approved):
Approved

PR #524, opened 2024-01-15
Refactor SAP transport and installation path handling

KimForss:
## Problem
The mount path FQDN was not correctly populated if private endpoints were not used

## Solution
Change the logic to use the public name of the resources 
## Tests

## Notes


hdamecharla (approved):
approved

PR #525, opened 2024-01-15
Use afs for shared storage

KimForss:
## Problem
ANF does not provide cross region resiliency, but Azure Files do

## Solution
Add a boolean flag 'use_AFS_for_shared_storage' that will cause the framework to create both the install and transport accounts using Azure Files but also create the ANF infrastructure for system storage

## Tests

## Notes


hdamecharla (approved):
Approved

PR #531, opened 2024-01-28
Provide a way to control SAS token access to storage accounts

KimForss:
## Problem

Customers who have policies enforcing AD authentication to storage accounts by blocking SAS token access cannot use SDAF

## Solution
Provide a way to control SAS token access to storage accounts.

## Tests

## Notes
The storage account that contains the SAP media must provide SAS token access as it will be accessed by the SAP Central Services VM which does not have Azure CLI available


hdamecharla (approved):
Approved

PR #532, opened 2024-01-29
Several DB2 HADR improvements

hjstam:
## Problem
1. Error: A local value with the name "sub_iscsi" has not been declared
2. The Azure scheduled events agent for pacemaker is installed by default for Red Hat clusters. We've seen unexpected behavior during our tests. For example pacemaker won't come online after the node has been fenced. The agent is als relatively new and we want to be able to test it thoroughly before implementing it on our production clusters. 
3. When using a custom NFS solution different pacemaker filesystem time-outs may be required.
4. DB2 HADR local and remote hostnames are set to system hostnames and not the virtual_hosts.
5. Tags in Azure for the anydb_node dbserver linux virtual machine are ignored while they are managed for the other linux virtual machines.
6. After a HADR fail-over to the stand-by node the application server can't connect due to a missing SSL certificate on the stand-by node.
7. Communication between primary and standby HADR nodes is unencrypted.

## Solution
1. `local.sub_iscsi.nsg.arm_id` should be `local.sub_iscsi_nsg_arm_id`
2. Add `cluster_use_scheduled_events_agent` boolean in ansible-input-api.yaml to manage Azure scheduled events agent installation. Default is set to true.
3. Add `custom_cluster_fs_mon_timeout` and `custom_cluster_sap_mon_timeout` variables (empty by default) in ansible-input-api.yaml for custom pacemaker NFS filesystem and SAP monitor timeouts. Refactored the cluster-Suse tasks so that it utilizes the calculated filesystem timeout and SAP resource monitor timeout from the 5.6.1-set_runtime_facts task.
4. Use virtual_host names for DB2 HADR local host name configuration.
5. Remove `tags` from ignore_changes for `azurerm_linux_virtual_machine.dbserver`
6. Generate a SSL certificate which is valid for 10 years for the db virtual hostname (loadbalancer), share the SSL keystore with the stand-by node and update DB2 SSL settings.
7. Encrypt communication between HADR Primary and Standby Instances with generated SSL certificate.

## Tests
Tested with a RHEL8.9 DB2 cluster.

## Notes
- https://community.sap.com/t5/technology-blogs-by-sap/encrypting-an-sap-system-on-a-db2-for-luw-database-reduce-downtime-by/ba-p/13530157
- https://www.sap.com/documents/2018/12/d2922a3b-307d-0010-87a3-c30de2ffd8ff.html

### DIFF
--- a/deploy/ansible/roles-db/4.2.1-db2-hainstall/tasks/4.2.1.4-db2_haparameters.yaml
+++ b/deploy/ansible/roles-db/4.2.1-db2-hainstall/tasks/4.2.1.4-db2_haparameters.yaml
@@ -54,6 +54,7 @@
             db2 update db cfg for {{ db_sid }} using HADR_REMOTE_HOST {{ db_hadr_remote_host }} immediate
             db2 update db cfg for {{ db_sid }} using HADR_REMOTE_SVC {{ db_sid | upper }}_HADR_2 immediate
             db2 update db cfg for {{ db_sid }} using HADR_REMOTE_INST db2{{ db_sid | lower }} immediate
+            db2 update db cfg for {{ db_sid }} using HADR_SSL_LABEL {{ db2_ssl_label }}
             db2 update db cfg for {{ db_sid }} using HADR_TIMEOUT 60 immediate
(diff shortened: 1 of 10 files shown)


KimForss (approved):
Looks good to me

PR #536, opened 2024-02-03
Feature : HANA Scale out with worker/stand by node using shared ANF storage

shekharsorot:
## Problem
This PR adds the HANA scale out with ANF feature to the main SDAF code base.
This fix is a bolt-on code that uses a single ansible variable "db_scale_out" to decide if the given 2 ( or more if composing sap-parameters.yaml by hand ) is going to be a tradational HSR multi site cluster or a single scale out + standby site.
Requires ANF volumes as per article : https://learn.microsoft.com/en-us/azure/sap/workloads/sap-hana-scale-out-standby-netapp-files-suse

Supports SLES and RHEl based hana scale out clusters.
Known issues / limitations.
1. Currently multi head network configuration is not implemented. All traffic will share the primary nic/ipconfig at the moment.
2. Single site setup only, no cross region or multi-site failover at the moment as it's not a certified architecture on Azure.
3. Minimum of 2 nodes are needed, second node is always configured as standby. It's expected that administrators update the second node to worker role when adding additional nodes as worker/standby.
4. All shared storage must be ANF or NetApp appliance-based exports (if not using TF to deploy workloads). Azure file share does not support the full subnet of NFS protocol due to which "hdblcm" will fail with an incorrect error message.
5. Hana scale out setup may take considerable amount of time depending on speed of underlying storage. This is normal as we start/stop Hana during the course of setup and this takes considerable time in Scale-Out systems.
6. Scale-Out - ANF requires enabling & setting Root login on all participating nodes. It's highly advised that post completion, administrators disable the Root account and re-configure /etc/ssh/sshd_config  to harden the system.
7. Firewall on RHEL & SLES systems is disabled to allow HANA scale out setup. Please refer to SAP documentation on port requirements if one wishes to re-enable firewall.


KimForss (approved):
Looks good to me

PR #537, opened 2024-02-05
Ensure mount the volumes before sap-notes

nnoaman:
## Problem
There's an issue related to compat-sap-c++-x, the link is being created to /usr/sap/lib/libstdc++.so.6 before mounting the volume, and this shouldn't be the case.

## Solution
Ensure mounting the volume before calling sap-notes role.

## Tests

## Notes


hdamecharla:
@nnoaman and @KimForss , let us have a quick call before we take a call on this PR.

PR #541, opened 2024-02-07
Ensure calling sap notes 3119751 after mounting the volume

nnoaman:
## Problem
There's an issue related to compat-sap-c++-*, the link is being created to /usr/sap/lib/libstdc++.so.6 before mounting the volume, and this shouldn't be the case.

## Solution
Calling sap notes 3119751 task after the mount

## Tests
Successfully tested after running the pipeline, ensuring that the link exists and points to the correct path.

## Notes


KimForss (approved):

KimForss:
Thanks @nnoaman

PR #544, opened 2024-02-08
Feature/scaleout-anf optimizations 

shekharsorot:
## Problem
currently we have duplicate code for ANF mount for HANA resulting due to having two different task files. This will result in potential mismatch of tasks in future when code is updates for non scale out or vice versa.

## Solution
HANA specific ANF Scale out tasks are now integrated into 2.6.1 and 2.6.8 with correct logic blocks. In the future, its intended to merge code for SAP ( tradational mount and simple-mount) as well to simplify code and reduce duplication

## Tests
This repo is a branch of azure/sap-automation:experimental. Standard pipeline execution should be run on SLES and RHEL based setup to verify that scale out setup is working as intended. for SLES, simple-mount option can be enabled to check if hana scaleout is working along side SAP application simple mount.


hdamecharla (approved):
Approved

PR #550, opened 2024-02-13
Suppress SCS install logs

hjstam:
## Problem
The SCS installation logs are always shown causing thousands off output lines even when the installation is successful.

## Solution
Like with the ERS installation the return code should be greater than zero.


KimForss (approved):

hdamecharla (approved):
Approved

PR #552, opened 2024-02-16
Only create/copy SSL cert when DB2 SSL communication is used

hjstam:
## Problem
SSL communication isn't enabled by default for DB2 ABAP installation. This can be managed by setting `db6.UseDb2SSLClientServerComm` to true in the BOM parameters file. Currently the code assumes that all DB2 ABAP installations use SSL Communication and that's the same keystore is used as for the DB2 encryption and that's not right.

## Solution
Check for the existence of the SSL and Encryption keystore and only import the tasks when they exist.

## Tests
- HA ABAP DB2 installation was completed successfully


hdamecharla (approved):
Approved

PR #553, opened 2024-02-20
Remove code to stop the secondary node in Oracle DB before restart

devanshjainms:
## Problem
<Please describe what problem this PR is trying to resolve>
During the testing of Oracle HA DB deployment, I found that the listener is not running on the secondary node, so stopping it would result in exception. Additionally, we are already performing the restart operation in the next step, so I think this step can be skipped.

## Solution
Skip stopping listener on secondary node.

## Tests
I performed 2 HA deployments with this change, and it was successful. 

## Notes


hdamecharla (approved):
Approved

PR #560, opened 2024-03-06
When Terraform plan failed, stop the script from executing Terraform apply

rdeveen:
## Problem
When the Terraform Plan has errors, the script will continue to run.

## Solution
The return_value was not being updated with the return code of the plan, so the check is not working.

## Tests
Before:
image

After:
image

## Notes


KimForss (approved):
Looks good to me

KimForss (approved):

PR #564, opened 2024-03-14
Hotfix for 3.11

KimForss:
## Problem
Fixing an issue which occurs when checking for the SAP Library storage account in a multi subscription scenario.

## Solution
Ensure that the check is done against the correct subscription

## Tests

## Notes


hdamecharla (approved):
Approved

PR #568, opened 2024-03-25
Hotfix/scaleout anf multi-node standby  (optional )

shekharsorot:
## Problem
current deployment of Scale out - ANF deploys a N+1 configuration where one node is always standby. This causes issues for customers where the on-prem/source HANA setup is 'N' node setup with no standby, causing HSR setup to fail.
Also, once HANA scale out is deployed, we cannot remove/change assigned roles to existing nodes, only new node addition can be done.

## Solution
Added a flag "database_HANA_no_standby_role" in terraform in addition to existing "database_HANA_use_ANF_scaleout_scenario" that will configure hdblcm setup to not add any standby role to any nodes during setup.
This variable adds an ansible variable "db_no_standby" that controls how "_rsp_additional_hosts" is passed to the HANA parameter file, i.e last node is worker/standby based on the value of this flag.

## Tests
Run a standard deployment with >2 nodes for HANA scale out - ANF based storage. Validate using python /usr/sap/<SID>/HDB<InstanceID>/exe/python_support/landscapeHostConfiguration.py post deployment of Scale out setup.

### DIFF
--- a/deploy/ansible/roles-db/4.0.3-hdb-install-scaleout/tasks/main.yaml
+++ b/deploy/ansible/roles-db/4.0.3-hdb-install-scaleout/tasks/main.yaml
@@ -7,7 +7,13 @@
 # |                  deploy hdblcm password file                               |
 # |                                                                            |
 # +------------------------------------4--------------------------------------*/
-
+# /*---------------------------------------------------------------------------8
+# |                                                                            |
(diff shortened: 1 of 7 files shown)


KimForss (approved):

KimForss:
Thank you @shekharsorot

PR #571, opened 2024-03-27
Control Private Endpoint DNS registration

KimForss:
## Problem
Currently Private Endpoints are automatically registered with DNS which causes problems for environments that manage this with policies,

## Solution
Add a variable 'register_endpoints_with_dns' that controls the Private Endpoint DNS registration. The variable defaults to true

## Tests

## Notes


hdamecharla (approved):
Approved #571

PR #574, opened 2024-04-01
Change to use import_role instead of include_role

KimForss:
## Problem
We don't have a good way to leverage tags to control task inclusion/exclusion

When using tags with **include_role**, the tags are applied only to the include_task itself - not the tasks inside the role!
When using tags with **import_role**, the tags are applied to all the tasks inside the role and not to the import_role task itself.

Reboots are not always working correctly using OracleLinux

## Solution

Change to use import_role instead of include_role

## Tests

## Notes


hdamecharla (approved):
Approved

PR #576, opened 2024-04-05
Bugfix - constraint Azure scheduled events agent installation

hjstam:
## Problem
cluster_use_scheduled_events_agent variable (default true) was added and is still in main. However the when statement disappeared - #532 

## Solution
Re-add when condition to constraint Azure scheduled events agent installation - #532


hdamecharla (approved):
Apporved.

PR #587, opened 2024-06-03
Custom LVM striping

SteffenBoThomsen:
## Problem

When using the extensibility of SDAF to create custom logical volumes, it's currentæy not possible to do LVM striping, which degrades performance of multi-pv LVMs.

## Solution
Use already established pattern from framework specific LVMs to define stripesize on custom logical volumes.The current implementation, ensures that we can do the following in our config repo tfvars file:

```
custom_logical_volumes = [
    {
      tier       = "sapos",
      node_tier  = "hana",
      vg         = "vg_data2",
      lv         = "lv_data2",
      stripesize = "{{ hana_data_stripe_size }}"
      size       = "100%FREE",
      fstype     = "xfs",
      path       = "/hana/data2"
    }
]
```

This will result in a stripesize similar to lv_hana_data. If stripesize is omitted, and the LVM consists of more than one pv, it will result in the `default_stripe_size` value. This has been added to the ansible-input-api.yaml file.

### DIFF
--- a/deploy/ansible/roles-os/1.5-disk-setup/tasks/1.5-custom-disks.yml
+++ b/deploy/ansible/roles-os/1.5-disk-setup/tasks/1.5-custom-disks.yml
@@ -47,6 +47,7 @@
     lv:                                "{{ item.lv }}"
     vg:                                "{{ item.vg }}"
     size:                              "{{ item.size }}"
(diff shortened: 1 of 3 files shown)


hdamecharla (approved):
Approved. Looks good to me.

PR #588, opened 2024-06-04
Perform az login before the az access token in ams provider creation

devanshjainms:
## Problem
While acquiring the access token using az cli command, an error occurred `Please run 'az login' to setup account.", "stderr_lines": ["ERROR: Please run 'az login' to setup account."], "stdout": "", "stdout_lines": []}
`
This behavior was not observed during the testing of the feature.

## Solution
Add az login task before acquiring the access token.


nnoaman (approved):
LGTM

PR #591, opened 2024-06-07
Various Ansible fixes

daradicscsaba:
## Problems

1. RHEL8.10 was released and the Framework doesn't support it

2. Cluster version check syntax is incorrect in roles  1.17-generic-pacemaker, 5.5-hanadb-pacemaker and 5.6-scsers-pacemaker, which also impedes a correct run of the installation process.

3. For DB2 deployments, the OS packages list contains libpam.so.0 and for this the i686 version of the pam package is a dependency. However, on RHEL8.9 VMs the x86_64 version of the pam package is also already present. Result is that 1.4-packages will error out for the DB2 nodes with:

```
  file /usr/share/man/man5/faillock.conf.5.gz from install of pam-1.3.1-33.el8.i686 conflicts with file from package pam-1.3.1-27.el8.x86_64
  file /usr/share/man/man5/limits.conf.5.gz from install of pam-1.3.1-33.el8.i686 conflicts with file from package pam-1.3.1-27.el8.x86_64
  file /usr/share/man/man5/namespace.conf.5.gz from install of pam-1.3.1-33.el8.i686 conflicts with file from package pam-1.3.1-27.el8.x86_64
...
  file /usr/share/man/man8/pam_usertype.8.gz from install of pam-1.3.1-33.el8.i686 conflicts with file from package pam-1.3.1-27.el8.x86_64
  file /usr/share/man/man8/pam_wheel.8.gz from install of pam-1.3.1-33.el8.i686 conflicts with file from package pam-1.3.1-27.el8.x86_64
  file /usr/share/man/man8/pwhistory_helper.8.gz from install of pam-1.3.1-33.el8.i686 conflicts with file from package pam-1.3.1-27.el8.x86_64
```

## Solutions

1. Variable added to the corresponding vars files to add support for RHEL8.10

2. Task that's setting the fact has been adjusted

3. To avoid the conflict, the existing x86_64 pam package needs to be updated first. Thus, it has been added to the package list with status 'latest', which is also now the first item in the loops  in deploy/ansible/roles-os/1.4-packages/tasks/1.4.1-packages.yaml. This results in the correct order of steps for this particular situation and should be a good mechanism for other similar conflicts.

## Tests

DB2 ABAP HA deployment

## Notes

Circumventing the pam conflict was also analyzed and tested with Red Hat by means of a case we opened.

Please incorporate these changes into a release ASAP because we need the RHEL8.10 support to create new systems.

### DIFF
--- a/deploy/ansible/roles-os/1.17-generic-pacemaker/tasks/1.17.2.0-cluster-RedHat.yml
+++ b/deploy/ansible/roles-os/1.17-generic-pacemaker/tasks/1.17.2.0-cluster-RedHat.yml
@@ -123,7 +123,7 @@
     - name:                            "1.17 Generic Pacemaker - Check if the pacemaker package version is greater than pacemaker-2.0.4"
       when:                            ansible_distribution_major_version in ["8", "9"]
       ansible.builtin.set_fact:
-        is_pcmk_ver_gt_204:            "{{ ansible_facts.packages['pacemaker'] is version('2.0.4', '>') | default(false) }}"
+        is_pcmk_ver_gt_204:            "{{ ansible_facts.packages['pacemaker'][0].version is version('2.0.4', '>') | default(false) }}"
 
     - name:                            "1.17 Generic Pacemaker - Ensure STONITH timeout is raised"
       ansible.builtin.command:         pcs property set stonith-timeout=900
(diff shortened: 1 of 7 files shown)


hdamecharla (approved):
Approved

PR #593, opened 2024-06-11
Set HDB schema name for ABAP and JAVA systems

jesperseverinsen:
## Problem
Current setup did not accommodate for JAVA HDB setups, failing trying to determine the HDB schema.

## Solution
Added a loop to determine schema name based on a list of property names, defaulting to `hana_schema` var if none is found. This has been pulled into a seperate `roles-db` role and used in the various locations, `schema_name` was set.


jesperseverinsen:
@microsoft-github-policy-service agree

hdamecharla (approved):
🥇 Approved

PR #598, opened 2024-06-19
Updates to scale out HSR code

shekharsorot:
1. Fixed an issue with RHEL/SUSE scale out failing when CIDR subnets are not provided entirely or in part.
2. Fixed an issue with RHEL based run failing when update packages is set to true
3. Fixed an issue where Python hook will fail to respond timely after completing Pacemaker configuration. ( Partially : Note we still need to find a way to prevent Role from moving to secondary site when we bring the cluster back up , which is causing replication to fail and undergo split brain )
4. Added code to ensure Majority maker ( if specified ) has necessary packages in place to support Pacemaker 
5. Added correct documentation links from Redhat documentation within code to allow devs to fetch correct cluster build procedure
6. Fixed an issue where global.ini will not get populated with correct interface IPs belonging to all the HANA hosts.
7. Fixed a race issue where using "pause" will result in one or more nodes in cluster configuration stage to race with each other, causing build to fail.
8. Imported changes from Experimental branch to ensure scaleout-hsr branch is upto-date with current bleeding edge code.
9. Fixed a small issue in configuration.sh script that could cause incorrect hosts file name generation
10. Ported changes from Windows BOM code to 3.3 BOM to allow for unpacking RAR based solid self extracting archives.
11. ( see above ) enabled EPEL for RedHat to install 'unar' to allow unpacking RAR archives from SAP ( refer to link https://access.redhat.com/solutions/28959 as to why unrar is no longer bundled in mainstream repository and distros anymore )

Todo in next PR:
1. Ensure pacemaker configuration always results in roles to be on primary site, ( once new sandbox azure environment is up and running )
2. Fix the issue with a proper logic where DBload fails due to incorrect DATA_UNITS path due to recent changes in archive format by SAP
3. Fix an issue where pacemaker does not clears azure_rst errors despite running resource cleanup. This needs low level debugging and requires new sandbox environment to be available.


hdamecharla (approved):
Approved

PR #604, opened 2024-07-22
Several (bug)fixes for RHEL deployments and deployments in general

hjstam:
## Problem
1. Fencing occurs during deployment and operation when SAP resources are stopped
2. systemd services aren't created on opposite SCS / ERS nodes
3. After installation (A)SCS resource group has a constraint because of the pacemaker fail-over which causes a problem during operation of the deployment
4. Bugfix single deployment where SCS node is also the NFS server. File systems are managed in the 2.6-sap-mounts role so the file systems aren't present when the 2.3-sap-exports role is executed and created directories are over-mounted.
5. The `sid_private_key` isn't required when the private key is available on the deployer by alternative means (SSH agent)
6. When using a per SID public key without a naming prefix the workload zone key vault secrets have conflicting names conflict `-sshkey / -sshkey-pub / -username` for every SID

## Solution
1. Add fast_stop=no to RHEL pacemaker filesystem resources to allow SAP (A)SCS / ERS to be stopped without killing and sapstartsrv starting processes again
2. Both (A)SCS and ERS systemd services should be available on both (A)SCS and ERS nodes
3. Clear the (A)SCS resource constraint as a post-install step
4. Remove directory / permission creation and let 2.3-sap-exports role only set-up the NFS server. In the 2.6-sap-mounts role the file systems are created and then are automatically available for the NFS server to be mounted on the other nodes which happens in the same role.
5. Make `sid_private_key` optional by wrapping it in a try statement
6. Add sdu_secret as resource prefix and suffix

## Tests
- Tested with ABAP DB2, Java DB2 and S4HANA single deployment where SCS is also the NFS server (without AFS/ANF)
- Tested with high available ABAP DB2, Java DB2 and S4HANA deployments

## Notes
- [What is the fast_stop option for a Filesystem resource in a Pacemaker cluster?](https://access.redhat.com/solutions/4801371)
- [Pacemaker: the configured stop timeout is not respected when stopping an OCFS2 filesystem resource]( https://www.suse.com/support/kb/doc/?id=000020860)


hjstam:
@hdamecharla please review this PR.

devanshjainms (approved):

PR #625, opened 2024-08-29
Configure constraint HANA ANF mounts only when HANA mountpoints are used

hjstam:
## Problem
When using ANF for the deployment without data, log and shared mountpoint on ANF the constraint shouldn't be configured. Otherwise HANA won't start.

## Solution
Check for valid hana_*_mountpoint variables.


KimForss (approved):
Thanks, good catch

PR #636, opened 2024-09-24
Enable setting custom local mountpoint options for HANA/DB2

daradicscsaba:
## Problem
Due to RHEL 8 security baseline requirements, we need to set custom mountpoint options for HANA and DB2 local filesystems and `/usr/sap`. Currently, the Framework doesn't support this, mountpoint options for said filesystems being hardcoded to `defaults`.

## Solution
Relevant tasks in role 2.6-sap-mounts have been adjusted, so that, when the variable `mntopts` is present in the disk config, the desired options are set, otherwise they default to `defaults`.

## Tests
Tested with non-HA HANA and DB2 ABAP deployments.

## Notes
none


KimForss (approved):
Looks good to me, thanks

PR #642, opened 2024-10-03
Align custom kernel parameters with extensibility documentation

SteffenBoThomsen:
## Problem
Currently the [extensibility documentation](https://learn.microsoft.com/en-us/azure/sap/automation/extensibility#adding-custom-kernel-parameters-linux) lists that customizations to kernel parameters can be done by

```yaml
custom_parameters:
  common:
    - { ... }
```

This is not the case as the code expects them grouped under the distribtuion_id.

## Solution

This PR ensures that both `common` and `distribution_id` can be utilized to set kernel parameters.
If both `common` and `distribution_id` sets the same property, the `distribution_id` will take precedence.
See the following example:

image

Which results in parameters being set like the following on the os level

image


devanshjainms (approved):

devanshjainms:
Thanks for the contribution. @SteffenBoThomsen

PR #665, opened 2024-11-26
Add two node cluster properties for DB2 on RHEL8

hjstam:
## Problem
Both nodes can be fenced when issues occur or you might end-up in a split-brain situation.

## Solution
Like with HANA and SCS/ERS we implement resource priority and set a fence delay of 15s.

## Notes
https://learn.microsoft.com/en-us/azure/sap/workloads/high-availability-guide-rhel-ibm-db2-luw?tabs=lb-portal isn't mentioning this (yet) but https://learn.microsoft.com/en-us/azure/sap/workloads/sap-hana-high-availability-rhel?tabs=lb-portal does.


hdamecharla (approved):
Approved #665

PR #668, opened 2024-11-29
Fix WL zone deployment by pipeline

ZsoltParoczi:
## Problem
The install_workloadzone.sh is not called by the 02 wlzone pipeline leveraging self hosted agent rather then deployer VM.

## Solution
Move the install_workloadzone.sh call out of the  if [ -f /etc/profile.d/deploy_server.sh ]; then

## Tests

## Notes


hdamecharla:
@ZsoltParoczi , we are working internally on an overhaul of the pipeline code. Some of the items mentioned here are already part of the new changes being made. Would request you to move this PR to draft state until you have a chance to review the new changes.

PR #670, opened 2024-12-04
Modify terraform dirs to be accessible without root user

ZsoltParoczi:
## Problem
The terraform init fails using self hosted agent with other than the root user.

## Solution
Change 
mkdir -p /opt/terraform/.terraform.d/plugin-cache
to
sudo mkdir -p /opt/terraform/.terraform.d/plugin-cache
in front 
of 
sudo chown -R $USER /opt/terraform

lines in the helper scripts.

## Tests

## Notes


ZsoltParoczi:
@microsoft-github-policy-service agree

PR #686, opened 2025-01-10
Release fixes

SteffenBoThomsen:
## Problem

This PR fixes two separate issues introduced by the latest SDAF release.

1. The introduction of new properties derived from the sap_namegenerator is not backwards compatible with the system removal pipeline when using the `{SID}_resource_names.json` to handle custom naming. This is due to the naming map being saved in the tfstate file and missing the new properties. Instead of having to modify the tfstate or doing a redeployment before removal, this PR ensures that the suffixes/prefixes that are missing from the object, are set to the defaults if they are missing.  

2. When running a pipeline on a branch, the current version correctly checks out the source code from the $(Build.SourceBranch), but utilizes the wrong environment variable for the branch inside the sh scripts. Meaning removal and adding of files to the config repo are always triggered on the main branch. This PR ensures that `$BUILD_SOURCEBRANCHNAME` is used instead of `$BRANCH`.

### DIFF
--- a/deploy/scripts/pipeline_scripts/01-control-plane-deploy.sh
+++ b/deploy/scripts/pipeline_scripts/01-control-plane-deploy.sh
@@ -112,8 +112,8 @@ fi
 
 cd "$CONFIG_REPO_PATH" || exit
 
-echo -e "$green--- Checkout $BRANCH ---$reset"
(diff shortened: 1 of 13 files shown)


KimForss (approved):
Looks good, thanks for doing this

PR #687, opened 2025-01-13
Fix subscription id persistance in mgmt key vault

SteffenBoThomsen:
## Problem

When deploying the workload zone, the environment subscription id is saved to the ControlPlane keyVault, in the newest release it wrongfully persists the `${STATE_SUBSCRIPTION}` instead of the workload zone `${subscription}`.


hdamecharla (approved):
Approved

PR #692, opened 2025-01-16
Set right uid as owner of /hana/* directories

SteffenBoThomsen:
## Problem
The newest release sets the `{{ sidadm }}` uid as owner of /hana/* directories. This user does not exist on the HANA VMs. It should instead be the ` {{ hdbadm}}` uid.


hdamecharla (approved):
approved

PR #733, opened 2025-03-28
Fix scale out v3.14.1

Rickmarges:
## Problem
HANA scale-out deployment was not working on RHEL 8.10

## Solution
See commits

## Tests
Manual RHEL 8.10 HANA scale-out deployment

## Notes


KimForss:
Hi @Rickmarges, thanks for this - a quick finding - for cases where you want to have more observers the observer zone variable should be a list so that you can distribute them across zones

Rickmarges:
Hi @KimForss, that only applies to anydb as with HANA it is either you deploy a observer or not (only 1 is possible).
Could you explain why having multiple observers are useful? And when not specifying the observer zone it defaults back to the previous behavior (using the db zones)

Rickmarges:
@KimForss I've now made it a list :) https://github.com/Azure/sap-automation/pull/733/commits/15a579361a67ad5763437ac4a335fcd13ca60f6b

KimForss:
Hi @Rickmarges The multiobserver scenario would be for Oracle where you may want to have an observer per zone.

KimForss (approved):

PR #778, opened 2025-07-03
2 HANA Scale out fixes for RHEL

Rickmarges:
## Problem
1. The sudo rule was a bit too strict for RHEL. The docs showed to include the `-t` and `-s` options, but a lot of 256 return codes showed up in the trace files.
2. For scale out the database loadbalancer would be connected to the admin subnet rather than the application subnet. This is only needed for ANF.

## Solution
1. Removed the `-t` and `-s` options from the sudo rules
2. Whenever a check on scaleout was done to use admin nic, a check for ANF usage was added


hdamecharla:
Hi @Rickmarges,

Thank you for submitting the pull request.

Could you please submit the changes to the following branch: https://github.com/Azure/sap-automation/tree/release/June-2025?

Let me know if you encounter any issues during the process.

hdamecharla:
Hi @Rickmarges,

Apologies for the earlier confusion. I have now switched to the `release/June-2025` branch and will proceed with the pull request review process accordingly.

PR #826, opened 2025-08-22
Fix role assignment conditions and update virtual network ID referenc…

devanshjainms:
This pull request updates the logic for assigning Azure role permissions and configuring DNS zone virtual network links in the Terraform modules for SAP deployment. The changes primarily improve how conditions are evaluated for resource creation and ensure the correct variables are used for virtual network IDs.

**Role assignment logic update:**

* Fixed the `count` condition in `azurerm_role_assignment.resource_group_user_access_admin_spn` to correctly assign the role when not running as MSI, ensuring the resource is only created when appropriate.

**DNS zone virtual network link improvements:**

* Updated the `count` condition for both `azurerm_private_dns_zone_virtual_network_link.vnet_mgmt_blob-agent` and `azurerm_private_dns_zone_virtual_network_link.vault_agent` to check the length of `additional_network_id` instead of using `contains` on keys, making the logic more robust and accurate. [[1]](diffhunk://#diff-7bb4132d1fed70f80eb44c9028a60c1f4d3a96f1a57339c9627f2ee9df029f07L87-R87) [[2]](diffhunk://#diff-7bb4132d1fed70f80eb44c9028a60c1f4d3a96f1a57339c9627f2ee9df029f07L141-R141)
* Changed the `virtual_network_id` assignments in both `vnet_mgmt_blob-agent` and `vault_agent` resources to use `var.dns_settings.additional_network_id` instead of `var.deployer_tfstate.additional_network_id`, ensuring the correct variable is referenced for the network ID. [[1]](diffhunk://#diff-7bb4132d1fed70f80eb44c9028a60c1f4d3a96f1a57339c9627f2ee9df029f07L106-R106) [[2]](diffhunk://#diff-7bb4132d1fed70f80eb44c9028a60c1f4d3a96f1a57339c9627f2ee9df029f07L160-R160)

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/infrastructure.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/infrastructure.tf
@@ -84,7 +84,7 @@ resource "azurerm_private_dns_zone_virtual_network_link" "vnet_mgmt_blob" {
 
 resource "azurerm_private_dns_zone_virtual_network_link" "vnet_mgmt_blob-agent" {
   provider                             = azurerm.dnsmanagement
-  count                                = var.dns_settings.register_storage_accounts_keyvaults_with_dns && length(var.dns_settings.additional_network_id) > 0 ? 1 : 0
+  count                                = var.dns_settings.register_storage_accounts_keyvaults_with_dns && contains(keys(var.deployer_tfstate), "additional_network_id") ? length(var.deployer_tfstate.additional_network_id) > 0 : false ? 1 : 0
   depends_on                           = [
                                            azurerm_storage_account.storage_tfstate,
                                            azurerm_private_dns_zone.blob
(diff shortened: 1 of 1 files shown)


deploy/terraform/terraform-units/modules/sap_library/infrastructure.tf
```diff
@@ -188,30 +188,3 @@ resource "azurerm_private_dns_zone_virtual_network_link" "vault_additional" {
   tags                                 = var.infrastructure.tags
 }
 
-
```
KimForss, line 191: This was here for the scenario where you don't have a deployer but want to add an extra network

devanshjainms: It was a duplicate of the resource on line 85

PR #844, opened 2025-09-10
Pipeline v1 fixes

SteffenBoThomsen:
Various small fixes to v1 pipelines:

1. In previous pipelines and version of SDAF it was possible to have multiple networks in the same workload zone, this fixes it across v1 pipelines to ensure that the framework is backwards compatible.
2. Extra line in post hook playbook call broke the post hook, removed the extra line to fix.
3. Ensure that system deployment correctly uses `VARIABLE_GROUP_ID`  instead of `VARIABLE_GROUP` in getVariableFromVariableGroup calls, to avoid int conversion error.


hdamecharla (approved):
Approved

PR #858, opened 2025-09-18
Oracle install fixes

SteffenBoThomsen:
Some fixes for oracle installations:

1. Ensure oracle-postprocessing pas in supported_tiers check is valid
2. ensure that the sap-automation provided installation param inifile works for oracle installations. The current solution ends up with multiple declarations of SAPINST.CD.PACKAGE.CD?, one for HANA and one for ORACLE, HANA is listed first and takes precedence. This is problematic when the default oracle installation extract the db export into {{ target_media_location }}/EXPORT and not {{ target_media_location }}/DB_EXPORT which is the case for HANA.


KimForss (approved):
Looks good to me

PR #859, opened 2025-09-19
Show the right state file in message.

SteffenBoThomsen:
## Problem

The success message almost gave me a heart attack

## Solution

Luckily it was just the message that needed updating


KimForss (approved):
:) Copy paste in action

hdamecharla (approved):
Approved

PR #869, opened 2025-09-24
 Update to the hosts.j2 Ansible template 

devanshjainms:
This pull request introduces a small logic improvement to the `hosts.j2` Ansible template to better handle virtual host names. The change ensures that the default virtual host is appended to the `virtual_host_names` list only when it is defined and different from the current host.

* Improved handling of `virtual_host_names` in `deploy/ansible/roles-sap-os/2.4-hosts-file/templates/hosts.j2` by conditionally appending the default virtual host only if it is set and not equal to the current host.


mkdeegan (approved):
Looks good and straight forward.

dhruvmicrosoft (approved):
looks good!

PR #871, opened 2025-09-26
fix: update version comparison for cloud-netconfig-azure in Pacemaker

devanshjainms:
This pull request updates the version comparison logic for the `cloud-netconfig-azure` package in Ansible playbooks to use the more robust `version` test instead of converting the version to a float. This change improves accuracy and reliability when checking package versions.

**Ansible playbook improvements:**

* Updated the version check for `cloud-netconfig-azure` in `1.17-generic-pacemaker` to use the `version` test for better reliability.
* Updated the version check for `cloud-netconfig-azure` in `1.18-scaleout-pacemaker` to use the `version` test for better reliability.


hdamecharla (approved):
Approved

PR #931, opened 2025-11-17
fix/kernel-param-ha-vm-swappiness

SteffenBoThomsen:
## Problem

Currentæy sap-automation-qa pipelines fail on configurations checks on systems deployed with SDAF, as the kernel parameter vm.swappiness does not have the right value.

## Solution

Set the common kernel parameter for vm.swappiness to 60 and for HA on SUSE set it to 10 according to the documentation here:
https://learn.microsoft.com/en-us/azure/sap/workloads/high-availability-guide-suse-pacemaker?tabs=msi#install-the-cluster


hdamecharla (approved):
Approved

nnoaman (approved):
LGTM

PR #1011, opened 2026-01-30
Fix: Resolve multiple bugs in setup WorkloadZone in ADO causing silent failures

### LINKED ISSUE #1009: [BUG]  New-SDAFADOWorkloadZone fail while reporting success and use wrong azure devops project to create variable groups
**Describe the bug**
Multiple bugs in `SDAFUtilities.psm1` cause the `New-SDAFADOWorkloadZone` cmdlet to reporting sucess while the end result is not achieved. These bugs affect Azure DevOps deployments where the default context doesn't match the target organization/project, caused variable groups created in the wrong project, and reporting success while there is an error while creating Service connection.

1. The `New-SDAFADOWorkloadZone` function (lines 1837-2416) never calls `az devops configure` to set the default context, unlike `New-SDAFADOProject` which does (line 1188).
2. Wrong Managed Identity Property used during service connection creation

**To reproduce**
1. Follow the powershell script in https://learn.microsoft.com/en-us/azure/sap/automation/configure-devops?tabs=linux#configure-azure-devops-services-for-sap-deployment-automation-framework
 to create two projects in two different organization, example, first SDAF-CTRL-IDCE in organization A then SAP Deployment in organization B. Note that conceptually we could use the same name in organization B, but in this case we didn't.
2. Follow the second powershell script  that calls New-SDAFADOWorkloadZone  with parameters referencing organization A and SDAF-CTRL-IDCE project.
3. Variable Groups were created in the SAP Deployment project, which has different organization and project name with the one specified in 2. User confused because checking the project and it has no variable group described by the output.
4. Error occured during service connection creation, but still reporting success

**Expected behavior**
The script should change the project context to the one specified on the parameter, or specify that the end user should change the project context first before running the cmdlet if the current context doesn't match the specified project and organization.
If the service connection creation failed then the script shouldn't report success.
The script should be adjusted to use correct properry so the service connection creation will succeeded.

**Additional context**

The source code : 

https://github.com/Azure/sap-automation/blob/main/deploy/scripts/pwsh/Output/SDAFUtilities/SDAFUtilities.psm1#L2199
**Line 2199:** Change:
```powershell
$ManagedIdentityClientId = $(az identity list --query "[?principalId=='$ManagedIdentityObjectId'].id" --subscription $ControlPlaneSubscriptionId --resource-group $ResourceGroupName --output tsv)
```
To:
```powershell
$ManagedIdentityClientId = $(az identity list --query "[?principalId=='$ManagedIdentityObjectId'].clientId" --subscription $ControlPlaneSubscriptionId --resource-group $ResourceGroupName --output tsv)
```

**Lines 1999-2000:** https://github.com/Azure/sap-automation/blob/main/deploy/scripts/pwsh/Output/SDAFUtilities/SDAFUtilities.psm1#L1999-L2000

This code blindly wrote that the Service connection created successfully.
```powershell
az devops service-endpoint create --service-endpoint-configuration $JsonInputFile --organization $AdoOrganization --project $AdoProject --output none --only-show-errors
Write-Host "Service connection '$ConnectionName' created successfully." -ForegroundColor Green
```

nnoaman:
## Summary
Fixes multiple bugs in `New-SDAFADOWorkloadZone` that caused false success reporting and resources created in wrong projects.

## Issues Addressed
- Variable groups created in wrong Azure DevOps project
- Service connection creation failing silently
- Wrong Managed Identity property causing authentication failures

### 1. Add Azure DevOps configuration
**Files:** `New-SDAFADOWorkloadZone.ps1`, `SDAFUtilities.psm1`

The function now calls `az devops configure --defaults` at the beginning to ensure all operations target the correct organization and project. This matches the pattern used in `New-SDAFADOProject`.

### 2. Fix Managed Identity Client ID query

Changed `.id` to `.clientId` in the az identity list query:
```diff
- az identity list --query "[?principalId=='...'].id"
+ az identity list --query "[?principalId=='...'].clientId"
```

The previous code returned the full ARM resource ID instead of the client ID required for service connection authentication.

### 3. Add error handling for service connection creation

Added $LASTEXITCODE checks after all az devops service-endpoint commands. The function now properly reports failures instead of displaying false success messages.

```diff 
  az devops service-endpoint create --service-endpoint-configuration $JsonInputFile --organization $AdoOrganization --project $AdoProject --output none --only-show-errors
+ if ($LASTEXITCODE -ne 0) {
+     Write-Error "Failed to create service connection"
+     throw "Service connection creation failed"
+ }
  Write-Host "Service connection '$ConnectionName' created successfully." -ForegroundColor Green
```

### DIFF
--- a/deploy/scripts/pwsh/Output/SDAFUtilities/SDAFUtilities.psm1
+++ b/deploy/scripts/pwsh/Output/SDAFUtilities/SDAFUtilities.psm1
@@ -1917,9 +1917,6 @@ function New-SDAFADOWorkloadZone {
     Write-Verbose "  CreateConnections: $CreateConnections"
     Write-Verbose "  ServiceManagementReference: $ServiceManagementReference"
 
-    Write-Verbose "Setting Azure DevOps defaults: organization=$AdoOrganization project=$AdoProject"
-    az devops configure --defaults organization=$AdoOrganization project="$AdoProject"
-
     # Initialize error tracking
     $ErrorActionPreference = 'Stop'
     $script:DeploymentErrors = @()
@@ -2183,6 +2180,9 @@ function New-SDAFADOWorkloadZone {
         throw "Project not found"
       }
 
+      Write-Verbose "Setting Azure DevOps defaults: organization=$AdoOrganization project=$AdoProject"
+      az devops configure --defaults organization=$AdoOrganization project="$AdoProject"
+
       $ControlPlaneVariableGroupId = (az pipelines variable-group list --query "[?name=='$ControlPlanePrefix'].id | [0]" --only-show-errors)
       $AgentPoolName = ""
       if ($ControlPlaneVariableGroupId.Length -ne 0) {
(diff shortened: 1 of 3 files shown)


hdamecharla (approved):
Approved

Issue #1007, opened 2026-01-28
Support region mapping indonesiacentral


yudhiwidyatama:
**Is your feature request related to a problem? Please describe.**
When deploying to region indonesiacentral, and we assume some region code in the WORKSPACE tfvars directory and file names, the region mapping process wil result region code UNKN. And the resource groups will be created with UNKN region code. Next when we run pipeline 'Remove deployments via ARM' it will try to destroy the region code from the directory and filenames (not UNKN) and will let the actual resource groups exist, resulting the removal to be ineffective.

**Describe the solution you'd like**
Microsoft could try to refactor / centralize the region mapping code so when there is a need to add a new region code it will only change one file / one mapping. Currently we have several files containing the region codes : 

```
boilerplate/02-sap-workload-zone.yml
deploy/scripts/pipeline_scripts/helper.sh
deploy/scripts/pipeline_scripts/22-sample-deployer-configuration.ps1
deploy/scripts/pipeline_scripts/v2/shared_functions.sh;;
deploy/scripts/pipeline_scripts/v2/shared_functions_v2.sh
deploy/scripts/deploy_utils.sh
deploy/pipelines/helper.sh
deploy/pipelines/02-sap-workload-zone.yaml
deploy/pipelines/variables/02-sap-workload-zone-variables.yaml
deploy/pipelines/22-sample-deployer-config.yaml

```

**Describe alternatives you've considered**
We could try to override the region code variable, but with current condition of many functionality duplication in sap-automation sources it is unclear whether any code will still picks from one of the 7 existing region code mapping in the source.

**Additional context**

yudhiwidyatama:
Hi, could we be informed what region code will be assigned to indonesiacentral? Or any due process that still in progress for the code assigment? Thanks in advance.

PR #1017, opened 2026-02-10
feat: add support for 'indonesiacentral' region in various configuration files and scripts

devanshjainms:

PR #1020, opened 2026-02-12
Add Region Code for Indonesia Central - Resolve https://github.com/Azure/sap-automation/issues/1007

yudhiwidyatama:
## Problem
Support region code mapping for region indonesia central
[ Github issue #1007 ](https://github.com/Azure/sap-automation/issues/1007)

## Solution
Add region code mapping for indonesia central.

## Tests
-

## Notes
If there are assigned region code for indonesia then the code IDCE assumed here need to be changed.


yudhiwidyatama:
Just read that issue #1017  referring to same issue and already has commits. Canceling this PR

PR #1128, opened 2026-06-03
Add pacemaker.timer configuration for delayed startup

devanshjainms:
This pull request introduces a new mechanism to control the startup timing of the Pacemaker service using a systemd timer, replacing previous autostart and delay logic. The main goal is to ensure Pacemaker starts after a configurable delay following system boot, improving cluster reliability and startup sequencing. The implementation includes new Ansible tasks, a template for the timer unit, and related variable configuration.

**Pacemaker startup delay and systemd timer integration:**

* Added a new Ansible task file (`1.17.2.1-pacemaker-timer.yml`) to deploy and configure a `pacemaker.timer` systemd unit, reload the systemd daemon, enable the timer, and disable Pacemaker and Corosync autostart. Also disables `SBD_DELAY_START` for certain stonith types, as the timer now handles the delay.
* Updated the main provisioning playbook (`1.17.2-provision.yml`) to include the new pacemaker timer configuration tasks.
* Introduced a new Jinja2 template (`pacemaker.timer.j2`) for the systemd timer unit, allowing the Pacemaker service to start after a configurable delay (`pacemaker_start_delay`).
* Defined the default `pacemaker_start_delay` variable in `main.yml`, with different defaults for SUSE and other OS families.

### DIFF
--- a/deploy/ansible/roles-os/1.17-generic-pacemaker/tasks/1.17.2-provision.yml
+++ b/deploy/ansible/roles-os/1.17-generic-pacemaker/tasks/1.17.2-provision.yml
@@ -89,3 +89,6 @@
 - name:                                "1.17 Generic Pacemaker - Cluster based on {{ ansible_os_family }} on VM {{ ansible_hostname }}"
   ansible.builtin.include_tasks:       "1.17.2.0-cluster-{{ ansible_os_family }}.yml"
   # when:                                cluster_existence_check != 0
+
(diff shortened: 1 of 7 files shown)


hdamecharla (commented):
Hi @devanshjainms , should we start the timer service as well? It would be nice to add a check to ensure the SBD config file is available.

KimForss:
@devanshjainms I would accept both the suggestions from CoPilot

KimForss (approved):

PR #1130, opened 2026-06-10
Remove Swift connection setup for web app and related role assignments

KimForss:
This pull request simplifies the `app_service.tf` Terraform module by removing the configuration for virtual network integration and some commented-out role assignments related to the web app. The main impact is that the web app will no longer be set up with a virtual network connection, and the unused or commented role assignment resources are cleaned up.

Key removals:

**Networking configuration:**
* Removed the `azurerm_app_service_virtual_network_swift_connection` resource, which previously set up VNet integration for the web app. This is not needed as the App Service is already attached to the subnet

**Role assignments cleanup:**
* Deleted commented-out `azurerm_role_assignment` resources for assigning the "Website Contributor" role to system-assigned and user-assigned identities, reducing clutter in the Terraform configuration.

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_deployer/app_service.tf
+++ b/deploy/terraform/terraform-units/modules/sap_deployer/app_service.tf
@@ -200,28 +200,3 @@ resource "azurerm_windows_web_app" "webapp" {
                                              }
 
 }
-
-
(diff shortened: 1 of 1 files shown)


nnoaman (approved):
LGTM!

PR #1223, opened 2026-07-29
Fix RHEL HANA HA/DR provider configuration and SBD prerequisites

devanshjainms:
## Summary
- Correct classic RHEL SAP HANA HA/DR provider paths and use the OS-aware provider-path variable.
- Align RHEL ANGI ChkSrv with `action_on_lost = fence`.
- Refactor iSCSI/SBD task configuration and add SBD prerequisites for SUSE and RHEL.

## Validation
- Parsed the updated HANA Pacemaker task files as YAML.
- Verified classic and ANGI provider paths, names, execution order, and ChkSrv action values locally.
- Ran `git diff --check`.


dhruvmicrosoft (approved):
lg!

PR #1288, opened 2026-08-26
3.23 Enhance deployment utilities, integrate SAP testing, and improve guidelines

devanshjainms:
This pull request introduces new AI-driven code review and marketplace plugin metadata for the SAP Deployment Automation Framework (SDAF), along with detailed review guidelines and reference documentation for key review dimensions. The main changes include the addition of plugin metadata files, expanded reviewer instructions, and comprehensive reference guides for correctness, domain, performance, and testing when reviewing infrastructure-as-code and automation scripts.

**Plugin metadata and marketplace integration:**

- Added `plugin.json` and `marketplace.json` files in both `.claude-plugin/` and `.github/plugin/` directories, defining the "azure-sap-automation" skillset and registering the SDAF plugin for marketplace discovery. These files describe the plugin, its skills, ownership, and relevant metadata. [[1]](diffhunk://#diff-037fda53525563fb89230704794a06db1150204ca28ce80697a637121c3ed3f9R1-R19) [[2]](diffhunk://#diff-1cf34e19fbf59e5d5d058f34580a7865a74c3d4f02550d8cb09832f691b9fcc6R1-R20) [[3]](diffhunk://#diff-5352ee4067f17e7948e1425843f0a454e045bc8edf338f0bcf75c6804ddabd9aR1-R16)

**Reviewer instructions and process:**

- Updated `.github/copilot-instructions.md` to direct reviewers to use the new `code-review` agent skill for reviewing Terraform, Ansible, shell, Python, and GitHub Actions workflow changes (excluding .NET code in `Webapp/`). The instructions clarify review priorities and the specific dimensions to be considered.

**Reference documentation for code review:**

- Added `.github/skills/code-review/references/correctness-and-terraform.md`, providing detailed guidance on correctness checks for Terraform code, including module structure, input wiring, resource identity, provider changes, and state semantics.
- Added `.github/skills/code-review/references/domain-performance-testing.md`, outlining review criteria for Azure/SAP domain rules, performance considerations, and testing coverage expectations for infrastructure-as-code and automation scripts.

PR #1292, opened 2026-08-27
Fix skill validation failures on release pull requests

devanshjainms:
## Problem

The skill validation workflow was failing on release pull requests for two unrelated reasons.

### 1. Fork pull requests can never run the evaluation

GitHub grants a pull request from a fork a **read-only** `GITHUB_TOKEN`. The `copilot-requests: write` permission the `evaluate` job requests is silently dropped, so the Copilot SDK cannot make a single model call and all 54 cases fail in ~2 seconds each.

Proof, from the job logs:

| | PR #1287 (fork `devanshjainms`), run 33020863371 | PR #1288 (same repo), run 33021010670 |
| --- | --- | --- |
| Granted | `Contents: read`, `Metadata: read` | `Contents: read`, **`CopilotRequests: write`**, `Metadata: read` |
| Secret source | `None` | `Actions` |
| Result | 54/54 fail | 53/54 pass |

This is not fixable on a fork pull request, so the job is now skipped there rather than reporting 54 false failures.

### 2. The `workspace-positive-02` prompt was defective

> How should I name the workspace and region/location portions of these parameter files?

The prompt has no SDAF anchor, `"these parameter files"` refers to nothing, and `"workspace"` reads as a generic Terraform workspace. The model answered about Terraform stacks and AWS regions and loaded no skill at all - **zero** invocations across all three attempts, consistently.

## Changes

- Skip `evaluate` on fork pull requests via `github.event.pull_request.head.repo.full_name == github.repository`.
- Reword `workspace-positive-02` using vocabulary `sdaf-workspace-and-tfvars` documents for itself (SDAF, WORKSPACES, region code, tfvars filename).
- Print a clear diagnostic when *every* case fails, so a credential or environment outage is no longer indistinguishable from a routing regression.

**No assertion is weakened.** Only the defective prompt changed; the skill documents were not edited to fit the test.

## Validation

- Live model run of the reworded prompt: **passes on the first attempt**, with exactly one `sdaf-workspace-and-tfvars` invocation, no stray companions, and the baseline correctly not invoking it.
- `pytest -o addopts="" tests/evals/test_evals.py` - 28 passed (26 before, +2 covering the new diagnostic).
- `pylint evals/evals.py` - 10.00/10.
- `black --check` - clean.
- Workflow YAML parsed and the guard expression asserted.

### DIFF
--- a/.github/workflows/skill-validation.yml
+++ b/.github/workflows/skill-validation.yml
@@ -161,6 +161,10 @@ jobs:
   evaluate:
     name: Evaluate skill routing
     needs: prepare
+    # Fork pull requests are granted a read-only GITHUB_TOKEN, so the
+    # `copilot-requests: write` permission requested below is silently dropped
+    # and every model call fails. Skip rather than report 54 false failures.
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     timeout-minutes: 120
     permissions:
--- a/evals/evals.json
+++ b/evals/evals.json
@@ -870,7 +870,7 @@
         },
         {
           "id": "workspace-positive-02",
-          "prompt": "How should I name the workspace and region/location portions of these parameter files?",
+          "prompt": "In the SDAF WORKSPACES tree, what is the naming convention for the workspace directory names and the region code portion of each tfvars filename?",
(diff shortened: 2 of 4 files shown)


hdamecharla (approved):
Approved

